PR #28, opened 2018-01-20
[index] Index deletes

hendrikmuhs:
Work in progress


hendrikmuhs:
Build failures due to too old gcc, looks like gcc 5 is required at minimum

PR #32, opened 2018-01-31
fix a bug in variable length short calculation which can cause corrupt keyvi files

hendrikmuhs:
fixes a off by 1 miscalculation between the length calculated by getVarshortLength and encodeVarshort, while the 1st returned 2 for 0x7fff the 2nd correctly returned 1 as 0x7fff fits in 1 byte (0111 1111 1111 1111 ).  The same bug existed for the next boundary 0x3fffffff. 

As a result of this miscalculation it was possible - although rather unlikely - that a bit in a bitvector was set, marking a written cell while it did not got written as the required space was 1 instead of 2. The consequence of the unwritten cell is a default to '\0', this '\0' caused a  mismatch starting with version 0.2 where '\0' became a valid transition. This broken transition caused a segfault  due to a broken pointer.

### DIFF
--- a/keyvi/include/keyvi/util/vint.h
+++ b/keyvi/include/keyvi/util/vint.h
@@ -106,7 +106,7 @@ size_t getVarintLength(int_t value) {
  */
 template <typename int_t = uint64_t>
 size_t getVarshortLength(int_t value) {
-  return (value > 0x1fffffffffff) ? 4 : (value < 0x3fffffff) ? (value < 0x7fff) ? 1 : 2 : 3;
+  return (value > 0x1fffffffffff) ? 4 : (value < 0x40000000) ? (value < 0x8000) ? 1 : 2 : 3;
 }
 
 /**
--- a/keyvi/include/keyvi/util/vint.h
+++ b/keyvi/include/keyvi/util/vint.h
@@ -59,7 +59,7 @@ void encodeVarint(int_t value, uint8_t* output, size_t* outputSizePtr) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


keyvi/tests/keyvi/util/vint_test.cpp
```diff
@@ -58,16 +58,33 @@ BOOST_AUTO_TEST_CASE(VShortLength) {
   BOOST_CHECK_EQUAL(77777, decodeVarshort(buffer));
 
   encodeVarshort(32767, buffer, &size);
-  BOOST_CHECK_EQUAL(util::getVarshortLength(1), size);
```
hendrikmuhs: Note: a unit test was in place but the check was rubbish. This should have been:

```
BOOST_CHECK_EQUAL(util::getVarshortLength(32767), size);
```

or 

```
BOOST_CHECK_EQUAL(1, size);
```

keyvi/include/keyvi/util/vint.h
```diff
@@ -106,7 +106,7 @@ size_t getVarintLength(int_t value) {
  */
 template <typename int_t = uint64_t>
 size_t getVarshortLength(int_t value) {
-  return (value > 0x1fffffffffff) ? 4 : (value < 0x3fffffff) ? (value < 0x7fff) ? 1 : 2 : 3;
+  return (value > 0x1fffffffffff) ? 4 : (value < 0x40000000) ? (value < 0x8000) ? 1 : 2 : 3;
```
narekgharibyan: Great!

PR #33, opened 2018-02-01
Test case for PR #32

narekgharibyan:
- added test case for PR #32 
- enabled integration-tests on Travis CI


hendrikmuhs:
LGTM

I never actually planned to execute it on travis because of resources required to run it. But as it works I am fine with it. Good stuff!

PR #35, opened 2018-02-04
[index] introduce subpackages

hendrikmuhs:
this PR shows how sub packages could look like.

These API changes would be available with 0.3 while the old imports into the main namespace would stay till 0.4.


hendrikmuhs:
@narekgharibyan let me know what you think about it

narekgharibyan:
It's great! 
I think later on we can add a sub-package called `keyvi.experimental`. Where we add experimental features, so they have some time to be tested, polished.

BTW, when you wanna have the 0.3 release ? What if we merge this sub-packaging PR to current master ?

hendrikmuhs:
@narekgharibyan 

This change is not compatible to master because the bigger restructuring part is not on master.

But having that said, I think we should consider merging the index branch.

PR #36, opened 2018-02-05
Tpie build dir

narekgharibyan:
Tpie build dir is now platform specific, so no need to manually delete build dirs when building let's say Mac and Manylinux versions locally.


hendrikmuhs:
I really like the idea. Unfortunately it will not cleanly merge with the index branch.

Would it be ok if we first merge 'index' and than re-implement this based on the merged master? FWIW: the python extension is now building keyvi as a whole, not just tpie, I will explain the reasons in the index branch landing PR.

narekgharibyan:
Yes, no problem.

narekgharibyan:
closing this in favor of #55

PR #37, opened 2018-02-05
Index Branch landing

hendrikmuhs:
This PR integrates the 'index' branch into master.

I think there are still some todo's left before finally merging. So the main purpose of this PR is to identify these todo's. Still those shall be the minimum viable things to fix before merging. 

'index' as a feature will remain 'experimental'. It is stable but has some rough edges, lacks some API's, documentation etc. But we can evolve that as part of master.

Todo's before merging:

- [x] finalize keyvimerger integration into python extension
- [x] investigate todo: 'check deleted keys in several segments' (fixed)

I will start documenting the design over the next days.

@narekgharibyan


keyvi/include/keyvi/dictionary/dictionary_merger.h
```diff
@@ -169,25 +180,30 @@ class DictionaryMerger final {
...
-      // get the weight value, for now simple: does not require access to the
-      // value store itself
-      handle.weight = value_store->GetMergeWeight(segment_it.entryIterator().GetValueId());
-
-      if (append_merge_) {
-        handle.value_idx =
-            value_store->GetMergeValueId(segment_it.segmentIndex(), segment_it.entryIterator().GetValueId());
+      if (!deleted_keys_[segment_it.segmentIndex()].empty() &&
+          top_key == deleted_keys_[segment_it.segmentIndex()].back()) {
+        // check the other deleted_keys,
```
hendrikmuhs: ^ todo

python/setup.py
```diff
@@ -214,37 +231,46 @@ def run(self):
...
+                                ' -D CMAKE_INSTALL_PREFIX={INSTALL_PREFIX}'.format(
+                CXX_FLAGS=CMAKE_CXX_FLAGS, INSTALL_PREFIX=keyvi_install_prefix)
+            if zlib_root is not None:
+                 keyvi_build_cmd += ' -D ZLIB_ROOT={ZLIB_ROOT}'.format(ZLIB_ROOT=zlib_root)
+            keyvi_build_cmd += ' ..'
+            keyvi_build_cmd += ' && make -j {}'.format(cpu_count)
+            keyvi_build_cmd += ' && make install'
+
+            print ("Building keyvi C++ part: " + keyvi_build_cmd)
+            subprocess.call(keyvi_build_cmd, shell=True)
```
hendrikmuhs: this is building keyvi now instead of only tpie, the reason is the keyvimerger binary

python/setup.py
```diff
@@ -214,37 +231,46 @@ def run(self):
...
+
+            # patch keyvimerger into the package
+            # note: package_data does not work for this as it would break sdist
+            build_dir = os.path.join(*([self.build_lib] + ['keyvi', '_bin']))
+            self.data_files.append(("keyvi._bin", keyvi_install_dir + "/bin", build_dir, ['keyvimerger']))
```
hendrikmuhs: ship keyvimerger as part of the package

python/src/addons/Index.pyx
```diff
@@ -0,0 +1,72 @@
...
+            params = args[1]
+        else:
+            raise Exception('can not handle type of %s' % (args,))
+
+        if isinstance(directory, unicode):
+            directory = directory.encode('UTF-8')
+
+        # inject keyvimerger
+        if not b"keyvimerger_bin" in params:
+            params[b"keyvimerger_bin"] = os.path.join(get_bin_folder(), b"keyvimerger")
```
hendrikmuhs: this allows to run keyvimerger from the python package, but the C++ side is using "keyvimerger_bin" yet (todo)

travis/build_linux.sh
```diff
@@ -5,6 +5,9 @@ mkdir build
 cd build
 cmake -DCMAKE_BUILD_TYPE=$CONF ..
 make -j 4
+
+sudo make install
```
hendrikmuhs: once the keyvimerger issue is fixed this can be reverted

keyvi/include/keyvi/index/internal/merge_job.h
```diff
@@ -0,0 +1,148 @@
...
+  size_t GetId() const { return id_; }
+
+  // todo: ability to kill job/process
+
+ private:
+  MergeJobPayload payload_;
+  size_t id_;
+  std::shared_ptr<TinyProcessLib::Process> external_process_;
+
+  void DoExternalProcessMerge() {
```
hendrikmuhs: so FWIW right now only 'external merging' is supported, this is wanted and an important aspect: merging can not crash the indexer, a OOM situation can be handled better (tbd: lower oom weight for merger) as well es priority can be set.

But: if the dictionaries are small, a in-process merge would be better. The idea in the long run is to let the merge policy decide about it.

python/setup.py
```diff
@@ -45,11 +58,11 @@ def symlink_keyvi():
     # workaround for autowrap bug (includes incompatible boost)
     autowrap_data_dir = "autowrap_includes"
 
-    dictionary_sources = path.abspath('keyvi')
-    tpie_build_dir = path.join(dictionary_sources, '3rdparty/tpie/build')
-    tpie_install_prefix = 'install'
-    tpie_include_dir = path.join(tpie_build_dir, tpie_install_prefix, 'include')
-    tpie_lib_dir = path.join(tpie_build_dir, tpie_install_prefix, 'lib')
+    dictionary_sources = path.abspath(keyvi_cpp_link)
+    keyvi_build_dir = path.join(keyvi_cpp, 'build')
```
hendrikmuhs: if I understood correctly this is where #36 should be implemented

keyvi/include/keyvi/dictionary/dictionary_merger.h
```diff
@@ -99,7 +121,8 @@ class DictionaryMerger final {
    *
    * @params params merger parameters
    */
-  explicit DictionaryMerger(const merger_param_t& params = merger_param_t()) : dicts_to_merge_(), params_(params) {
+  explicit DictionaryMerger(const keyvi::util::parameters_t& params = keyvi::util::parameters_t())
```
narekgharibyan: Probably it worth to add `using keyvi::util::parameters_t` on class/function level and then just use `parameters_t` as a type in order to have less dense code.

keyvi/include/keyvi/dictionary/dictionary_merger.h
```diff
@@ -197,20 +243,69 @@ class DictionaryMerger final {
...
   std::string manifest_ = std::string();
+  MergeStats stats_;
+
+  /**
+   * Load a file with deleted keys if it exists
+   */
+  std::vector<std::string> TryLoadDeletedKeys(const std::string& filename) {
+    std::vector<std::string> deleted_keys;
+    boost::filesystem::path deleted_keys_file{filename};
+    deleted_keys_file += ".dk";
```
narekgharibyan: Looks like this naming assumption is only for `Index`, maybe it worth to have merger more independent and make this configurable or provided from outside ? (no strong opinion though)

keyvi/include/keyvi/dictionary/dictionary_merger.h
```diff
@@ -197,20 +243,69 @@ class DictionaryMerger final {
...
+    if (deleted_keys_stream.good()) {
+      TRACE("found deleted keys file");
+
+      {
+        // reads the buffer as 1 big chunk, could be improved
+        // msgpack v2.x provides a better interface (visitor)
+        std::stringstream buffer;
+        buffer << deleted_keys_stream.rdbuf();
+
+        msgpack::unpacked unpacked_object;
```
narekgharibyan: Also the format of file is defined somewhere else, will be great that writer and reader parts would live together in centralized place.

python/src/py/keyvi/compiler/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+'''
+
+from keyvi._core import CompletionDictionaryCompiler, CompletionDictionaryMerger, IntDictionaryCompiler, IntDictionaryMerger
+from keyvi._core import JsonDictionaryCompiler, JsonDictionaryCompilerSmallData, JsonDictionaryMerger
+from keyvi._core import KeyOnlyDictionaryCompiler, KeyOnlyDictionaryGenerator, KeyOnlyDictionaryMerger
+from keyvi._core import StringDictionaryCompiler, StringDictionaryMerger
```
narekgharibyan: Will be good to have empty new line added in the EOF.

keyvi/include/keyvi/index/internal/base_index_reader.h
```diff
@@ -0,0 +1,85 @@
...
+template <class PayloadT, class SegmentT = ReadOnlySegment>
+class BaseIndexReader {
+ public:
+  using const_segments_t = const std::shared_ptr<std::vector<std::shared_ptr<SegmentT>>>;
+
+  template <typename... Args>
+  explicit BaseIndexReader(Args... args) : payload_(args...) {}
+
+  dictionary::Match operator[](const std::string& key) {
+    dictionary::Match m;
```
narekgharibyan: Will be better to have a full variable name

keyvi/include/keyvi/index/internal/merge_job.h
hendrikmuhs: well, hopefully merger is stable enough to be safe from crashes, compiler is also internal, so if there is a bug we have a problem anyhow. If an indexer process crashes it can be restarted and it will continue from where it stopped but we loose all data that were in internal buffers (like the current compiler). This problem is true for any data store. 

At some point in the future I would like to provide a transaction log, that can be plugged. But even then, you need an fsync to be sure that data is backed up.

For this version, data is guaranteed to be backed to disk after calling Flush(), this triggers a compile of the latest data including an implicit fsync at the end.

keyvi/include/keyvi/dictionary/dictionary_merger.h
hendrikmuhs: yeah, I thought about an explicit parameter in the merger. On the other side you have to create a file called 'keyvifile.kv.dk' - rather unlikely to do this by mistake.

Thoughts?

keyvi/include/keyvi/index/internal/merge_job.h
narekgharibyan: Agree, but in that case if we are "okay" for this kind of bugs to crash it, and then we restart the index. Shouldn't we be okay for just spawning a merger thread and making sure that we are catching all exceptions from that thread ?

--

Yes of course this makes it less safe, but looks like it can be a good tradeoff.

hendrikmuhs: the 'internal merger' will spawn a thread and do the merge as part of the thread, it is important that `merge_job` does not block. The interface will be the same so that it is an internal detail whether the merge happens in-process or external.

I haven't started creating safe-guards yet, till now I preferred to let it crash, e.g. the return code isn't checked yet. It's on the list. ;-)

keyvi/include/keyvi/dictionary/dictionary_merger.h
narekgharibyan: Fair enough.

keyvi/include/keyvi/util/single_producer_consumer_ringbuffer.h
```diff
@@ -22,89 +22,79 @@
...
-      }
+    if (next_head == tail_.load(boost::memory_order_acquire)) {
       // spin-lock
-      while (next_head == tail_.load(boost::memory_order_acquire))
-      {
+      while (next_head == tail_.load(boost::memory_order_acquire)) {
         std::this_thread::sleep_for(std::chrono::milliseconds(SPINLOCK_WAIT_IN_MS));
       }
     }
     ring_[head] = std::move(value);
```
narekgharibyan: As the value is `const`, `std::move()` has no effect.

keyvi/include/keyvi/util/single_producer_consumer_ringbuffer.h
```diff
@@ -22,89 +22,79 @@
...
     head_.store(next_head, boost::memory_order_release);
-    return true;
+    return;
```
narekgharibyan: `return` statement is not needed

python/setup.py
```diff
@@ -214,37 +233,41 @@ def run(self):
...
+            CMAKE_CXX_FLAGS = '-fPIC -std=c++11'
+            if sys.platform == 'darwin':
+                CMAKE_CXX_FLAGS += ' -mmacosx-version-min=10.9'
 
-                CMAKE_CXX_FLAGS = '-fPIC -std=c++11'
-                if sys.platform == 'darwin':
-                    CMAKE_CXX_FLAGS += ' -mmacosx-version-min=10.9'
+            keyvi_build_cmd = 'mkdir -p {}'.format(keyvi_build_dir)
+            keyvi_build_cmd += ' && cd {}'.format(keyvi_build_dir)
+            keyvi_build_cmd += ' && cmake -D CMAKE_BUILD_TYPE:STRING=bindings ' \
```
narekgharibyan: As `CMAKE_BUILD_TYPE` is usually used to specify build type, e.g. Release, Debug and etc. I'd suggest using `cmake --build` command with `--target` option. Also we will not need the IF-ELSE clause and TPIE workaround in CMakeLists.txt file. 

I was facing the same issue for Rust: https://github.com/narekgharibyan/keyvi-1/commit/385d61de9939e4ff7dd248ee5d4bfa93f66d753f

python/setup.py
narekgharibyan: Yes, will redo #38 once this PR is merged.

python/setup.py
```diff
@@ -194,13 +210,16 @@ class bdist(custom_opts, _bdist.bdist):
...
+
         class bdist_wheel(custom_opts, _bdist_wheel.bdist_wheel):
             parent = _bdist_wheel.bdist_wheel
             user_options = _bdist_wheel.bdist_wheel.user_options + custom_user_options
+
         have_wheel = True
     except: None
 
-    class build_ext(_build_ext.build_ext):
+    class build_py(_build_py.build_py):
```
narekgharibyan: Maybe I'm mistaken, but looks like we don't need to derive from`build_py` anymore, and `build_ext` only is sufficient.

python/tests/index/merger_binary_test.py
```diff
@@ -0,0 +1,32 @@
...
+def get_package_root():
+    module_location = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(keyvi._pycore.__file__)), ".."))
+
+    if (sys.version_info.major >= 3):
+        module_location = module_location.encode('utf-8')
+
+    return module_location
+
+
+def get_interpreter_executable():
```
narekgharibyan: Maybe I'm missing something, but as in `Index.pyx` we are using `get_interpreter_executable()` and `get_package_root()` from `keyvi._pycore`. It seems it's better to use them here (and have tested) as well. 

Maybe even have a function returning `get_interpreter_executable() + b" " + os.path.join(get_package_root(), b"_pycore" , b"keyvimerger.py")`, so we are sure that tests and merger are utilizing same code.

@hendrikmuhs what do you think about ?

hendrikmuhs: I was thinking that both are internal functions, not available from external.

But I managed to find them, they are in keyvi._core, so yes,makes sense, I will change it.

python/setup.py
hendrikmuhs: @narekgharibyan 

the best alternative I can come up with:

```
add_custom_target(bindings
    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:tpie> ${CMAKE_BINARY_DIR}
    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:tiny-process-library> ${CMAKE_BINARY_DIR}
    DEPENDS tpie tiny-process-library
)
```

that create a target bindings (`make bindings`) and copies both libraries into the cmake build root. Otherwise it is hard to get the libs and a `make install` always triggers building everything.

narekgharibyan: I see, but probably this is a better alternative.

keyvi/include/keyvi/index/internal/segment.h
```diff
@@ -0,0 +1,206 @@
...
+        new_delete_(false),
+        deleted_keys_swap_filename_(path) {
+    if (load) {
+      Load();
+    }
+
+    deleted_keys_swap_filename_ += ".dk-swap";
+  }
+
+  explicit Segment(const boost::filesystem::path& path, const std::vector<std::shared_ptr<Segment>> parent_segments,
```
narekgharibyan: Maybe this should be `const std::vector<std::shared_ptr<Segment>> &` ?

python/src/addons/ReadOnlyIndex.pyx
```diff
@@ -0,0 +1,36 @@
+
+
+    def get (self, key, default = None):
```
narekgharibyan: I see that we use lowercase `get()` to have `dict` like interface, but I've noticed for `Dictionary()` users often times it's quite confusion that all methods are uppercase except one. Maybe it's worth to have `Get()` here ?

narekgharibyan:
@hendrikmuhs Looks great!!

As you have mentioned maybe there are some rough edges, but I think those will be identified better after release, when it's in use for some time.

PR #38, opened 2018-02-06
[index] finish keyvimerger integration as part of the python package

hendrikmuhs:
with this PR the python package is fully self-contained again. keyvimerger already shipped with the package but the code in the C++ part was missing.


hendrikmuhs:
removing `make install` broke C++ unit tests. But I consider this a good thing. Tests should run without running `make install` first, i somehow need to inject the keyvimerger binary location like for the python tests.

hendrikmuhs:
to be reviewed as part of #37

PR #39, opened 2018-02-12
Fuzzy match

narekgharibyan:
- implemented fuzzy matching
 - added Dictionary::GetFuzzy() method
 - fixed use of .clang-format file
 - added ConstructorInitializerIndentWidth: 4 to .clang-format

Note: At the moment Python bindings are missing, will add them in a separate PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ FILE(GLOB_RECURSE UNIT_TEST_SOURCES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} keyvi/t
 add_executable(keyvicompiler keyvi/bin/keyvicompiler/keyvicompiler.cpp)
 add_executable(keyviinspector keyvi/bin/keyviinspector/keyviinspector.cpp)
 add_executable(keyvimerger keyvi/bin/keyvimerger/keyvimerger.cpp)
-add_executable(unit_test_all ${UNIT_TEST_SOURCES} keyvi/include/keyvi/dictionary/fuzzy_match.h)
+add_executable(unit_test_all ${UNIT_TEST_SOURCES})
 
 target_link_libraries(keyvicompiler tpie ${Boost_LIBRARIES} ${ZLIB_LIBRARIES})
 target_link_libraries(keyviinspector tpie ${Boost_LIBRARIES} ${ZLIB_LIBRARIES})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ FILE(GLOB_RECURSE UNIT_TEST_SOURCES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} keyvi/t
 add_executable(keyvicompiler keyvi/bin/keyvicompiler/keyvicompiler.cpp)
 add_executable(keyviinspector keyvi/bin/keyviinspector/keyviinspector.cpp)
 add_executable(keyvimerger keyvi/bin/keyvimerger/keyvimerger.cpp)
-add_executable(unit_test_all ${UNIT_TEST_SOURCES})
+add_executable(unit_test_all ${UNIT_TEST_SOURCES} keyvi/include/keyvi/dictionary/fuzzy_match.h)
 
 target_link_libraries(keyvicompiler tpie ${Boost_LIBRARIES} ${ZLIB_LIBRARIES})
 target_link_libraries(keyviinspector tpie ${Boost_LIBRARIES} ${ZLIB_LIBRARIES})
--- a/keyvi/check-style.sh
+++ b/keyvi/check-style.sh
@@ -8,7 +8,7 @@ cpplint_files=()
 for file in ${infiles}; do
   fqfile=`git rev-parse --show-toplevel`/${file}
   echo "Checking: ${file}"
-  if ! cmp -s ${fqfile} <(clang-format ${fqfile}); then
+  if ! cmp -s ${fqfile} <(clang-format -style=file ${fqfile}); then
     clang_format_files+=("${file}")
   fi
   if ! cpplint --quiet ${fqfile}; then
--- a/keyvi/check-style.sh
+++ b/keyvi/check-style.sh
@@ -8,7 +8,7 @@ cpplint_files=()
 for file in ${infiles}; do
   fqfile=`git rev-parse --show-toplevel`/${file}
   echo "Checking: ${file}"
-  if ! cmp -s ${fqfile} <(clang-format -style=file ${fqfile}); then
+  if ! cmp -s ${fqfile} <(clang-format ${fqfile}); then
     clang_format_files+=("${file}")
   fi
   if ! cpplint --quiet ${fqfile}; then
--- a/keyvi/include/keyvi/dictionary/dictionary.h
+++ b/keyvi/include/keyvi/dictionary/dictionary.h
@@ -32,9 +32,9 @@
 #include "dictionary/fsa/automata.h"
 #include "dictionary/fsa/state_traverser.h"
 #include "dictionary/fsa/traverser_types.h"
-#include "dictionary/fuzzy_match.h"
 #include "dictionary/match.h"
 #include "dictionary/match_iterator.h"
+#include "dictionary/matching/fuzzy_matching.h"
 
 // #define ENABLE_TRACING
 #include "dictionary/util/trace.h"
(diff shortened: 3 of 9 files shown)


keyvi/check-style.sh
```diff
@@ -8,7 +8,7 @@ cpplint_files=()
 for file in ${infiles}; do
   fqfile=`git rev-parse --show-toplevel`/${file}
   echo "Checking: ${file}"
-  if ! cmp -s ${fqfile} <(clang-format ${fqfile}); then
+  if ! cmp -s ${fqfile} <(clang-format -style=file ${fqfile}); then
```
hendrikmuhs: can you explain, is that something that does not work on your setup?

CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ FILE(GLOB_RECURSE UNIT_TEST_SOURCES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} keyvi/t
 add_executable(keyvicompiler keyvi/bin/keyvicompiler/keyvicompiler.cpp)
 add_executable(keyviinspector keyvi/bin/keyviinspector/keyviinspector.cpp)
 add_executable(keyvimerger keyvi/bin/keyvimerger/keyvimerger.cpp)
-add_executable(unit_test_all ${UNIT_TEST_SOURCES})
+add_executable(unit_test_all ${UNIT_TEST_SOURCES} keyvi/include/keyvi/dictionary/fuzzy_match.h)
```
hendrikmuhs: this should not be required

keyvi/check-style.sh
narekgharibyan: At some point `clang-format` was not running, and it seemed to me that this had fixed it.
But just tested, it's fine even without it. Thx!

CMakeLists.txt
narekgharibyan: done

keyvi/include/keyvi/dictionary/dictionary.h
```diff
@@ -32,6 +32,7 @@
 #include "dictionary/fsa/automata.h"
 #include "dictionary/fsa/state_traverser.h"
 #include "dictionary/fsa/traverser_types.h"
+#include "dictionary/fuzzy_match.h"
```
hendrikmuhs: I find it a bit confusing: 

- fuzzy_match.h sounds like a derived version of match.h, but that isn't the case.
- it is place top-level, do you think it will be used directly or always via dictionary?

keyvi/include/keyvi/dictionary/dictionary.h
```diff
@@ -381,6 +382,12 @@ class Dictionary final {
     return MatchIterator::MakeIteratorPair(tfunc);
   }
 
+  MatchIterator::MatchIteratorPair GetFuzzy(const std::string& query, size_t max_edit_distance) {
```
hendrikmuhs: could that be const?

keyvi/include/keyvi/dictionary/fuzzy_match.h
```diff
@@ -0,0 +1,131 @@
...
+#include "dictionary/match.h"
+#include "dictionary/util/utf8_utils.h"
+
+namespace keyvi {
+namespace dictionary {
+
+class FuzzyMatch {
+ public:
+  FuzzyMatch(fsa::automata_t fsa, const std::string& query, size_t max_edit_distance)
+      : fsa_(std::move(fsa)), query_(query), max_edit_distance_(max_edit_distance) {
```
hendrikmuhs: I think std::move is not effective here

keyvi/include/keyvi/dictionary/fuzzy_match.h
```diff
@@ -0,0 +1,131 @@
...
+#include "dictionary/fsa/automata.h"
+#include "dictionary/fsa/codepoint_state_traverser.h"
+#include "dictionary/fsa/traverser_types.h"
+#include "dictionary/match.h"
+#include "dictionary/util/utf8_utils.h"
+
+namespace keyvi {
+namespace dictionary {
+
+class FuzzyMatch {
```
hendrikmuhs: final?

(at least until we find a reason to derive something from it)

keyvi/include/keyvi/dictionary/fuzzy_match.h
hendrikmuhs: I would also init first_match_ as part of the constructor

keyvi/include/keyvi/dictionary/fuzzy_match.h
```diff
@@ -0,0 +1,131 @@
...
+ public:
+  FuzzyMatch(fsa::automata_t fsa, const std::string& query, size_t max_edit_distance)
+      : fsa_(std::move(fsa)), query_(query), max_edit_distance_(max_edit_distance) {
+    std::vector<int> codepoints;
+    utf8::unchecked::utf8to32(query.begin(), query.end(), back_inserter(codepoints));
+
+    query_char_length_ = codepoints.size();
+    metric_ptr_.reset(new stringdistance::Levenshtein(codepoints, 20, max_edit_distance_));
```
hendrikmuhs: I suggest to move that down after the prefix has been matched, so in case prefix does not match we avoid initialization cost

keyvi/include/keyvi/dictionary/fuzzy_match.h
```diff
@@ -0,0 +1,131 @@
...
+  fsa::automata_t fsa_;
+  std::string query_;
+  const size_t max_edit_distance_;
+
+  size_t exact_prefix_in_chars_ = 0;
+  size_t query_char_length_ = 0;
+
+  Match first_match_;
+
+  std::unique_ptr<stringdistance::Levenshtein> metric_ptr_;
```
hendrikmuhs: that looks quite nice, so it would be easy to turn that into 

template<class MetricT = stringdistance::Levenshtein>
class FuzzyMatch final { 
...
};

Not saying we need that now.

keyvi/include/keyvi/dictionary/dictionary.h
hendrikmuhs: what about fuzzy_matcher.h placed in a 'matching' subfolder?

In the long run I would migrate e.g. GetNear there as well. Yours look cleaner than my in-class implementations.

keyvi/tests/keyvi/dictionary/fuzzy_match_test.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+#include <boost/test/unit_test.hpp>
+
+#include "dictionary/dictionary.h"
+#include "testing/temp_dictionary.h"
+
+namespace keyvi {
+namespace dictionary {
+
+BOOST_AUTO_TEST_SUITE(FuzzyMatchTests)
```
hendrikmuhs: could you add some test cases for edge cases:

- no match
- empty string
- query shorter than prefix length

some CJK would be nice as well, although you use umlauts.

hendrikmuhs: some tests for the score would be nice as well. E.g. when it comes to non-ascii it is important that e.g. a mismatch of 'ü' scores edit distance 1 (codepoint length) and not 2 (byte length).

hendrikmuhs:
added some comments. Looks pretty good!

keyvi/tests/keyvi/dictionary/fuzzy_match_test.cpp
narekgharibyan: Sure!

keyvi/include/keyvi/dictionary/dictionary.h
narekgharibyan: Yup!

keyvi/include/keyvi/dictionary/dictionary.h
narekgharibyan: Yes, truth be told I was not sure about naming and location either. But I wanted to have it outside of Dictionary class for sure!
Also during testing I've realized that current implementation of `FuzzyCompletion` has a couple of bugs, and as it shares most of code with `FuzzyMatch`, later on it can be reimplemented by just deriving from `FuzzyMatch` and overriding `NextMatch()` method.

keyvi/include/keyvi/dictionary/fuzzy_match.h
narekgharibyan: Sure thing!

keyvi/include/keyvi/dictionary/dictionary.h
hendrikmuhs: that sounds like a good idea, but again we have to differ between the matching and the completion case, still a lot of code can be shared.

FWIW 

 - fuzzy completion/matching isn't well tested, I agree
 - it has been written before I rewrote lots of the traversal code (https://github.com/KeyviDev/keyvi/blob/master/keyvi/include/keyvi/dictionary/fsa/traversal/traversal_base.h), which can make a big difference in speed
 - bugs: sure, please open issues for them, I really like to improve the quality of fuzzy matching

PR #41, opened 2018-02-16
[index] instrument merger with stats and make interface more like compiler

hendrikmuhs:
this change instruments dictionary merger to retrieve stats about number of deleted/overwritten keys

### DIFF
--- a/keyvi/include/keyvi/dictionary/dictionary_merger.h
+++ b/keyvi/include/keyvi/dictionary/dictionary_merger.h
@@ -57,17 +57,15 @@ struct merger_exception : public std::runtime_error {
   using std::runtime_error::runtime_error;
 };
 
+struct MergeStats {
+  size_t number_of_keys_ = 0;
+  size_t deleted_keys_ = 0;
+  size_t updated_keys_ = 0;
+};
+
 template <class PersistenceT, class ValueStoreT = fsa::internal::NullValueStore>
 class DictionaryMerger final {
   using GeneratorAdapter = fsa::GeneratorAdapterInterface<PersistenceT, ValueStoreT>;
-
- public:
-  struct MergeStats {
-    MergeStats() : number_of_keys_(0), deleted_keys_(0), updated_keys_(0) {}
-    size_t number_of_keys_;
-    size_t deleted_keys_;
-    size_t updated_keys_;
-  };
 
  private:
   class SegmentIterator {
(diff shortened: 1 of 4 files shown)


keyvi/include/keyvi/dictionary/dictionary_merger.h
```diff
@@ -49,8 +49,26 @@
...
 template <class PersistenceT, class ValueStoreT = fsa::internal::NullValueStore>
 class DictionaryMerger final {
+  using GeneratorAdapter = fsa::GeneratorAdapterInterface<PersistenceT, ValueStoreT>;
+
+ public:
+  struct MergeStats {
+    MergeStats() : number_of_keys_(0), deleted_keys_(0), updated_keys_(0) {}
+    size_t number_of_keys_;
+    size_t deleted_keys_;
+    size_t updated_keys_;
```
narekgharibyan: As we are defining `default constructor` which only initializes members with zeros, we can use `default member initializers`  instead.
```
size_t number_of_keys_ = 0;
```

keyvi/include/keyvi/dictionary/dictionary_merger.h
```diff
@@ -220,21 +245,37 @@ class DictionaryMerger final {
...
-    generator->WriteToFile(filename);
+  void Write(std::ostream& stream) {
+    if (!generator_) {
+      throw merger_exception("not merged yet");
+    }
+
+    generator_->Write(stream);
   }
 
+  template <typename StringType>
```
narekgharibyan: Do you a scenario that this will be different from `std::string` ? 
It fells like that we don't need the `template` here.

hendrikmuhs: good point, that allows usage of "const char*" but there is no need for such super-flexibility. Let's stick to std::string.

FWIW I will change it in the other places as well.

narekgharibyan: `const char*` anyways will be converted to `std::string` by additional constructor call, which I think is affordable.

hendrikmuhs:
Some explanation for that change:

This goes beyond 'useful information' but has the following usecase: An 'append merge' - which only concatenates value stores - is much cheaper than a real merge. With instrumenting the merger and being able to get the stats about merging, the caller can track 'wasted space' in the value store.

E.g. the segment merger in the index can decide to do 'append merges' until certain thresholds are reached and then do a real merge to free up the wasted space.

FWIW I decided against letting the merger embed the stats. The caller can do that via manifest.

narekgharibyan:
Sounds cool!

narekgharibyan:
LGTM

PR #43, opened 2018-02-17
Minor type fixes and formatting

narekgharibyan:
This PR include small type fixes from https://github.com/cliqz-oss/keyvi/pull/242
As `style-check` requires whole file to be formatted properly decided to have a separate PR for this, to make review easier.

### DIFF
--- a/keyvi/CPPLINT.cfg
+++ b/keyvi/CPPLINT.cfg
@@ -1,2 +1,3 @@
 linelength=120
 root=keyvi/include
+filter=-build/include_subdir
--- a/keyvi/include/keyvi/dictionary/completion/multiword_completion.h
+++ b/keyvi/include/keyvi/dictionary/completion/multiword_completion.h
@@ -22,20 +22,23 @@
  *      Author: hendrik
  */
 
-#ifndef MULTIWORD_COMPLETION_H_
-#define MULTIWORD_COMPLETION_H_
+#ifndef KEYVI_DICTIONARY_COMPLETION_MULTIWORD_COMPLETION_H_
+#define KEYVI_DICTIONARY_COMPLETION_MULTIWORD_COMPLETION_H_
 
-#include "stringdistance/levenshtein.h"
-#include "utf8.h"
+#include <string>
+#include <vector>
+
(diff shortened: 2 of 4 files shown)


keyvi/CPPLINT.cfg
```diff
@@ -1,2 +1,3 @@
 linelength=120
 root=keyvi/include
+filter=-build/include_subdir
```
hendrikmuhs: can you explain, afaik that enforces always using full qualified directories. Not to bad.

hendrikmuhs:
@narekgharibyan 

as far as I remember the only non-formatting change was the size_t fixes, right?

keyvi/CPPLINT.cfg
narekgharibyan: Agree! 
But as we include `#include "utf8.h"` from CMake include paths, and there it's not hierarchy, I couldn't do `#include "utf8/utf8.h"`

The alternative could be to change `CMakeLists.txt` to have `include_directories(keyvi/3rdparty)` 

Or to use includes from `utf8.h` directly. 
```
#include "utf8/checked.h"
#include "utf8/unchecked.h"
```
Didn't have strong opinion which is the best, so decided to send a PR and we discuss.

narekgharibyan:
Yes, but as I already had the code changed/formatted locally. Thought to send a PR, not to spend a time in the future.

keyvi/CPPLINT.cfg
narekgharibyan: FYI: the same change was needed in https://github.com/KeyviDev/keyvi/pull/44 as well.

hendrikmuhs: ah I see, so it is because the 3rdparty dependency isn't integrated like tpie, rapidjson and the others.

this is from utf8cpp, origin http://utfcpp.sourceforge.net/ (currently offline)

there are several github versions, e.g. https://github.com/ledger/utfcpp - does not look official.

So the alternative would be to fix the 3rdparty paths, introducing subdirectories there, but the same problem applies to md5 and tiny-process.

I suggest: keep this fix for later.

PR #44, opened 2018-02-17
C Api and Rust bindings

narekgharibyan:
- Keyvi C Api which can be utilized later on to add bindings to other languages (e.g. Go)
 - Keyvi Rust bindings,
 - Rust builds on Travis CI
 - Now style-check doesn't force short functions implemented outside of class to be merged to one line.

@hendrikmuhs This PR is same as https://github.com/cliqz-oss/keyvi/pull/242  with addressed review comments. Also Rust package was renamed from `rskeyvi` to `keyvi`.


rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "keyvi"
+version = "0.0.1"
```
hendrikmuhs: I get the semantics behind that version number. It would still make sense to have a unified version number.

rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+build = "build.rs"
+authors = ["Narek Gharibyan <narekgharibyan@gmail.com>", "Subu <subu@cliqz.com>"]
+
+[build-dependencies]
+bindgen = "0.30.0"
+cmake = "*"
```
hendrikmuhs: can you say something like > 2.8.12 (that's the minimum right now)?

rust/Cargo.lock
```diff
@@ -0,0 +1,549 @@
+[root]
+name = "keyvi"
+version = "0.0.1"
```
hendrikmuhs: version needs to be set twice (it is set in the ".toml" file below as well)?

rust/Cargo.lock
```diff
@@ -0,0 +1,549 @@
...
+dependencies = [
+ "bindgen 0.30.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cmake 0.1.25 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.18 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rayon 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
```
hendrikmuhs: are these dependencies? Quite a lot, is that correct?

hendrikmuhs:
Look great!

I wonder a bit about:

- version number, we should use the same version everywhere, why does it have to be set twice?
- dependencies, I do not know much about rust, but that looks really strange to me
- a Readme.md file in the rust folder with some basic getting started documentation would be nice (can be added later)

I would like to merge the index branch first in case of merge trouble. You introduced some style fixes - which are good - but if we merge this first I might have to reformat all files.

narekgharibyan:
- version number: Sure, that makes sense! It can be that in the future we will have a bug/fix in bindings part only, which I think will require patch version change and rerelease only for a particular binding. But that should be okay. 
- dependencies: I removed `Cargo.lock` file from VCS, as it is not needed for libraries, and is required only for executables, also added `[dev-dependencies]` section for test requirements. Now distribution package is dependent only on 3 packages.
- Readme.md: Sure, need to add it! Probably along with release on https://crates.io/

Agree, I'll do the review next week, so we can first merge Index and then Rust and make the `0.3` release.

PR #45, opened 2018-02-18
[index] fix chmod of keyvimerger binary

hendrikmuhs:
the original implementation turned out to be broken. This change fixes keyvimerger access rights and adds a test for it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.0.2)
 project(keyvi)
 
 string(TOLOWER ${CMAKE_BUILD_TYPE} LOWERCASE_CMAKE_BUILD_TYPE)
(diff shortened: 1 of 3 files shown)


hendrikmuhs:
After some more testing it seems that the current approach will not work for wheel packages. Wheels have no execution hook but simply unzip the package and changer permissions (The wheel still has '+x' on the binary, but after installation it is reset). 

Therefore I need a different approach, possible solutions:

 - create a python version and install it as script and/or use keyvi cli (problem: different style of cmd arguments)
 - at runtime copy the binary to a tmp location with proper access rights and use that
 - provide a script that needs to be run after installation that fixes rights (could be in combination with the above as fallback)
 - implement in-process merge and only use that as a fallback

hendrikmuhs:
It goes into the right direction:

Providing the binary works, some build fail due to full static linkage:

- on manylinux we lack a static build of boost
- on mac full static linkage is not possible, but I need to link it with minimal dependencies as we already do for the python bindings

Both is complicated to fix, maybe I still go with the option of a python script.

narekgharibyan:
I think `python script` is better option. Also we will not need to build whole `keyvi` from `setup.py` and ship with package, but just `extension` part.

hendrikmuhs:
closing in favor of #48

PR #51, opened 2018-03-04
fix a couple of type issues in minimization hash

hendrikmuhs:
fix a couple of namings and types in minimization hash as well as memory usage calculation

### DIFF
--- a/keyvi/tests/keyvi/dictionary/fsa/internal/minimization_hash_test.cpp
+++ b/keyvi/tests/keyvi/dictionary/fsa/internal/minimization_hash_test.cpp
@@ -69,6 +69,19 @@ BOOST_AUTO_TEST_CASE(reset) {
   delete hash;
 }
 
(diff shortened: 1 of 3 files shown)


keyvi/tests/keyvi/dictionary/fsa/internal/minimization_hash_test.cpp
```diff
@@ -69,6 +69,19 @@ BOOST_AUTO_TEST_CASE(reset) {
   delete hash;
 }
 
+BOOST_AUTO_TEST_CASE(memory_usage) {
+  // default
+  MinimizationHash<PackedState<>> *hash = new MinimizationHash<PackedState<>>();
+  BOOST_CHECK_EQUAL(hash->GetMemoryUsage(), (8171 + 2042) * sizeof(PackedState<>));
+  delete hash;
```
narekgharibyan: Maybe we can allocate object in stack, or use `std::unique_ptr` here ?

PR #56, opened 2018-03-16
fixed JumpConsistentHashString() bindings for Python3, Python2 unicode

narekgharibyan:
- fixed JumpConsistentHashString() bindings for Python3, Python2 unicode
- suppressing stdout from merger binary test


hendrikmuhs:
Looks good as is, but there is also an alternative:

If we change JumpConsistentHashString to take a std::string as input instead of a C string, you could use libcpp_utf8_string for the binding, not requiring to write extra code.

narekgharibyan:
@hendrikmuhs thanks for the hint, this way it looks much better!

PR #57, opened 2018-03-20
Return Unicode string from keyvi python API

narekgharibyan:
- fixed unit tests to be compatible with new unicode format
 - added OutputStdStringUnicodeConverter for return string types


python/src/converters/__init__.py
```diff
@@ -3,3 +3,4 @@
 
 def register_converters():
     special_converters.append(MatchIteratorPairConverter())
+    special_converters.append(OutputStdStringUnicodeConverter())
```
hendrikmuhs: we should try to get this into autowrap.

Having that said, as we do not know if and when they accept it, this is of course fine for now.

python/tests/test_tools.py
```diff
@@ -20,12 +20,4 @@ def tmp_dictionary(compiler, file_name):
     os.remove(fq_file_name)
 
 def decode_to_unicode(data):
```
hendrikmuhs: can we remove it or does it require a lot more changes?

python/src/converters/__init__.py
narekgharibyan: Sure, will send a PR to them as well.

PR #59, opened 2018-03-20
keyvi vector

narekgharibyan:
This implements `keyvi.vector` which is a `vector` like `read-only` data-structure implemented by utilizing `mmap`. It enables possibility to access data via just 2 page accesses and basically has a 0 memory requirement.

Some stuff that can be added later: 
 - `msgpack` can be used from python bindings to make it even faster
 - `resize()` like functionality in order to enable assignment by index, and not just only `PushBack`
 - `cross-architecture` support - `done`

### DIFF
--- a/keyvi/include/keyvi/vector/vector_file.h
+++ b/keyvi/include/keyvi/vector/vector_file.h
@@ -79,7 +79,8 @@ class VectorFile {
     const auto map_options = dictionary::fsa::internal::MemoryMapFlags::FSAGetMemoryMapOptions(loading_strategy);
     index_region_ = mapped_region(file_mapping, boost::interprocess::read_only, in_stream.tellg(), index_size, nullptr,
                                   map_options);
-    index_region_.advise(mapped_region::advice_types::advice_random);
+    const auto advise = dictionary::fsa::internal::MemoryMapFlags::FSAGetMemoryMapAdvices(loading_strategy);
+    index_region_.advise(advise);
 
     in_stream.seekg(size_t(in_stream.tellg()) + index_size);
     value_store_reader_.reset(dictionary::fsa::internal::ValueStoreFactory::MakeReader(
--- a/keyvi/include/keyvi/vector/vector_file.h
+++ b/keyvi/include/keyvi/vector/vector_file.h
@@ -1,0 +1,147 @@
+/* * keyvi - A key value store.
+ *
+ * Copyright 2018   Narek Gharibyan<narekgharibyan@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ *  vector_file.h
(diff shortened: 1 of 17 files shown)


keyvi/include/keyvi/vector/vector_file.h
```diff
@@ -0,0 +1,147 @@
...
+#include "dictionary/fsa/internal/value_store_factory.h"
+#include "util/serialization_utils.h"
+#include "vector/types.h"
+
+static const char KEYVI_VECTOR_BEGIN[] = "KEYVI_VECTOR_BEGIN";
+static const size_t KEYVI_VECTOR_BEGIN_LEN = 18;
+static const char KEYVI_VECTOR_END[] = "KEYVI_VECTOR_END";
+static const size_t KEYVI_VECTOR_END_LEN = 16;
+static const char MANIFEST[] = "manifest";
+static const char SIZE[] = "size";
```
hendrikmuhs: as this is a header implementation, I suggest to be careful on naming, SIZE and MANIFEST sound very generic.

keyvi/include/keyvi/vector/vector_file.h
```diff
@@ -0,0 +1,147 @@
...
+
+    const auto index_ptree = util::SerializationUtils::ReadJsonRecord(in_stream);
+    size_ = boost::lexical_cast<size_t>(index_ptree.get<std::string>(SIZE));
+    const auto index_size = size_ * sizeof(offset_type);
+
+    auto file_mapping = boost::interprocess::file_mapping(filename.c_str(), boost::interprocess::read_only);
+    const auto map_options = dictionary::fsa::internal::MemoryMapFlags::FSAGetMemoryMapOptions(loading_strategy);
+    index_region_ = mapped_region(file_mapping, boost::interprocess::read_only, in_stream.tellg(), index_size, nullptr,
+                                  map_options);
+    index_region_.advise(mapped_region::advice_types::advice_random);
```
hendrikmuhs: this is not make use of loading strategy, you could use FSAGetMemoryMapAdvices(loading_strategy)

keyvi/include/keyvi/vector/vector_generator.h
```diff
@@ -0,0 +1,91 @@
...
+
+  VectorGenerator(const VectorGenerator&) = delete;
+
+  ~VectorGenerator() { boost::filesystem::remove_all(temporary_directory_); }
+
+  void PushBack(typename ValueStoreT::value_t value) {
+    bool dummy_minimization = false;
+    const offset_type value_idx = value_store_->GetValue(value, &dummy_minimization);
+
+    index_store_->Append(&value_idx, sizeof(offset_type));
```
hendrikmuhs: it would be good to make this platform independent. This is easily done by defining your preferred endianess, I would go for little endian, which is Intel's default. 

You need to:

include endian.h

call htole64(v) for host to persistence order and le64toh(v) for persistence to host order on lookup. On Intel the macros are optimized out.

keyvi/tests/keyvi/vector/basic_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+  StringVectorGenerator generator;
+
+  const std::string filename = "vector_string.kvv";
+  const size_t size = 100;
+
+  for (size_t i = 0; i < size; ++i) {
+    generator.PushBack(std::to_string(i));
+  }
+
+  generator.WriteToFile(filename);
```
hendrikmuhs: would be good to cleanup files afterwards, even better if you use TMP, e.g. something like:

```
boost::filesystem::path temp_path = boost::filesystem::temp_directory_path();

temp_path /= boost::filesystem::unique_path("vector-unit-test-string-test-%%%%-%%%%-%%%%-%%%%");
std::string file_name = temp_path.string();

...

std::remove(file_name)
```

Issue #60, opened 2018-03-20
index can crash if running out of file handles


hendrikmuhs:
With the default number of file handles (1024) on an ubuntu 16.04 the indexer can crash caused by a corrupted file.

This seems to originate from an unhandled problem on a filestream somewhere in the value store/memory map manager. The error gets ignored, the output file gets corrupted and the indexer crashes shortly after  trying to read the file.

Increasing the number of file handles workarounds the problem.

A fix should

- improve error handling, properly detect if we run out of file handles
- throttle indexer to avoid problems if file handles get short
- proper error messages

hendrikmuhs:
Not yet fixed but:

- #61 improves error messages, so it gets visible if we run out of file handles
- #62 fixes a lazy loading problem found while debugging as well as limits the number of files for 1 merge making, so that it is less likely to run out of file handles

Next steps:

- try to increase number of file handles at startup
- limit number of segments proportional to the number of filehandles per process and throttle indexing if there are to many segments

PR #61, opened 2018-03-23
add error handling in memorymapmanager

hendrikmuhs:
adds error handling to memory map manager so that problems like exhausted file handles get visible.

relates #60

### DIFF
--- a/keyvi/include/keyvi/dictionary/fsa/internal/memory_map_manager.h
+++ b/keyvi/include/keyvi/dictionary/fsa/internal/memory_map_manager.h
@@ -42,6 +42,14 @@ namespace dictionary {
 namespace fsa {
 namespace internal {
 
+/**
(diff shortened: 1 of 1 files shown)


keyvi/include/keyvi/dictionary/fsa/internal/memory_map_manager.h
```diff
@@ -164,10 +172,19 @@ class MemoryMapManager final {
...
+          throw memory_map_manager_exception("failed to open data stream");
+        }
+
+        stream << data_stream.rdbuf();
+
+        if (!data_stream) {
+          throw memory_map_manager_exception("failed to write into output stream");
+        }
+
+        data_stream.close();
```
narekgharibyan: associated file will automatically be closed when stream object goes out of the scope

PR #64, opened 2018-03-27
re-add loading_strategy_types

hendrikmuhs:


python/src/py/keyvi/dictionary/__init__.py
```diff
@@ -17,4 +17,4 @@
 limitations under the License.
 '''
 
-from keyvi._core import Dictionary
+from keyvi._core import Dictionary, loading_strategy_types
```
narekgharibyan: @hendrikmuhs I expect later on `keyvi.vector` will support `loading_strategy_types` as well, maybe it's worth to expose it on `global` level only ?

hendrikmuhs: ok, makes sense

PR #77, opened 2018-04-25
Introduce -Wall flag, and fix compiler warnings

amit-cliqz:
Fixed  compiler warnings and formatted those files which required fixes.


keyvi/include/keyvi/dictionary/completion/prefix_completion.h
```diff
@@ -173,7 +173,7 @@ class PrefixCompletion final {
           if (data->traverser) {
             TRACE("Current depth %d", exact_prefix + data->traverser.GetDepth() - 1);
 
-            int score =
+            size_t score =
```
hendrikmuhs: a score isn't a size type. What's the problem with int? I am open for replacing it with int32_t, to make it more explicit as int isn't clearly defined. A 64bit type would be to large, given that for the matrix calculation it would mean a massive size increase.

keyvi/tests/keyvi/dictionary/fsa/internal/memory_map_manager_test.cpp
```diff
@@ -23,33 +23,31 @@
  *      Author: hendrik
  */
 
-#include <cstring>
-#include <boost/test/unit_test.hpp>
-
 #include "dictionary/fsa/internal/memory_map_manager.h"
+#include <boost/test/unit_test.hpp>
+#include <cstring>
```
hendrikmuhs: clang-format does not get it right, the order should be:

- C includes
- empty line
- C++ includes
- empty line
- boost includes
- empty line
- other 3rdparty includes
- empty line
- keyvi includes except for trace
- empty line
- trace include, with comment for enabling it

amit-cliqz: Thanks Hendrik for the information. Is it possible that clang-format behaves differently on different platform? I see one of the builds is failing if I correct the order. How should I proceed here if the clang-format doesn't accept the aforementioned order.

keyvi/include/keyvi/dictionary/matching/fuzzy_matching.h
```diff
@@ -93,7 +93,7 @@ class FuzzyMatching final {
         continue;
       }
 
-      if (traverser_ptr_->IsFinalState() && metric_ptr_->GetScore() <= max_edit_distance_) {
+      if (traverser_ptr_->IsFinalState() && metric_ptr_->GetScore() <= static_cast<int>(max_edit_distance_)) {
```
narekgharibyan: @amit-cliqz do we still need this type casting ?

amit-cliqz: @narekgharibyan yes, unless we refactor Distance matrix and GetScore function. I have a TODO for that.

keyvi/include/keyvi/dictionary/completion/prefix_completion.h
```diff
@@ -173,6 +173,7 @@ class PrefixCompletion final {
           if (data->traverser) {
             TRACE("Current depth %d", exact_prefix + data->traverser.GetDepth() - 1);
 
+            // TODO(Amit) use int32_t instead of int.
```
narekgharibyan: Can't we do the similar change to `GetScore()`, by just moving your `TODO` and type cast inside `Put()` method, and use `size_t` already here ?

keyvi/tests/keyvi/dictionary/fsa/internal/memory_map_manager_test.cpp
```diff
@@ -33,23 +34,22 @@ namespace dictionary {
...
   boost::filesystem::path path = boost::filesystem::temp_directory_path();
-  path /= boost::filesystem::unique_path(
-      "dictionary-fsa-unittest-%%%%-%%%%-%%%%-%%%%");
+  path /= boost::filesystem::unique_path("dictionary-fsa-unittest-%%%%-%%%%-%%%%-%%%%");
   boost::filesystem::create_directory(path);
   MemoryMapManager m(chunkSize, path, "basic test");
 
   m.GetAddress(0);
-  char* testptr = (char*) m.GetAddress(2 * chunkSize);
+  char* testptr = reinterpret_cast<char*>(m.GetAddress(2 * chunkSize));
```
narekgharibyan: Maybe I'm mistaken, but do we need `reinterpret_cast` ? I assume `static_cast` should be enough.

amit-cliqz: Yes, static_cast should be enough here.

PR #79, opened 2018-04-29
Index Autoconfiguration, overcoming filedescriptor limitations

### LINKED ISSUE #60: index can crash if running out of file handles

hendrikmuhs:
This change implements auto-configuration of index settings, including configuring the maximum number of segments which is directly connected to the number of file descriptors. If not externally given it tries to increase the limit (usually from 1024 to 1M). In case of running low on file descriptors index operations are automatically throttled so that concurrent compaction can catch up and reduce open descriptors.

Also changes the queue implementation - after benchmarking - to use condition variables instead of a spin-lock if the queue runs full or empty. This improves indexing performance and also fixes a problem in the spin-lock implementation that could cause cpu hogging. 

fixes #60

### DIFF
--- a/keyvi/include/keyvi/index/constants.h
+++ b/keyvi/include/keyvi/index/constants.h
@@ -43,4 +43,7 @@ static const char DEFAULT_KEYVIMERGER_BIN[] = "keyvimerger";
 // spinlock wait time if there are to many segments
 static const size_t SPINLOCK_WAIT_FOR_SEGMENT_MERGES_MS = 10;
 
+// max parallel process for segment merging
+static const size_t MAX_CONCURRENT_MERGES_DEFAULT = 8;
+
 #endif  // KEYVI_INDEX_CONSTANTS_H_
--- a/keyvi/include/keyvi/index/constants.h
+++ b/keyvi/include/keyvi/index/constants.h
@@ -40,7 +40,7 @@ static const size_t DEFAULT_REFRESH_INTERVAL = 1000ul;
 static const size_t DEFAULT_COMPILE_KEY_THRESHOLD = 10000ul;
 static const char DEFAULT_KEYVIMERGER_BIN[] = "keyvimerger";
 
-// spinlock wait time if there are to many segments
+// spinlock wait time if there are too many segments
 static const size_t SPINLOCK_WAIT_FOR_SEGMENT_MERGES_MS = 10;
 
 // max parallel process for segment merging
--- a/keyvi/include/keyvi/index/constants.h
+++ b/keyvi/include/keyvi/index/constants.h
@@ -41,6 +41,6 @@ static const size_t DEFAULT_COMPILE_KEY_THRESHOLD = 10000ul;
 static const char DEFAULT_KEYVIMERGER_BIN[] = "keyvimerger";
 
 // spinlock wait time if there are to many segments
-static const size_t SPINLOCK_WAIT_FOR_SEGMENT_MERGES = 10;
+static const size_t SPINLOCK_WAIT_FOR_SEGMENT_MERGES_MS = 10;
 
 #endif  // KEYVI_INDEX_CONSTANTS_H_
--- a/keyvi/include/keyvi/index/constants.h
+++ b/keyvi/include/keyvi/index/constants.h
@@ -25,11 +25,22 @@
 #ifndef KEYVI_INDEX_CONSTANTS_H_
 #define KEYVI_INDEX_CONSTANTS_H_
 
+#include <cstddef>
+
 static const char INDEX_REFRESH_INTERVAL[] = "refresh_interval";
 static const char MERGE_POLICY[] = "merge_policy";
 static const char DEFAULT_MERGE_POLICY[] = "simple";
 static const char KEYVIMERGER_BIN[] = "keyvimerger_bin";
 static const char INDEX_MAX_SEGMENTS[] = "max_segments";
 static const char SEGMENT_COMPILE_KEY_THRESHOLD[] = "segment_compile_key_threshold";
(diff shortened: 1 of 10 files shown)


keyvi/include/keyvi/index/constants.h
```diff
@@ -25,9 +25,25 @@
...
+static const char INDEX_MAX_SEGMENTS[] = "max_segments";
+static const char SEGMENT_COMPILE_KEY_THRESHOLD[] = "segment_compile_key_threshold";
+static const char MAX_CONCURRENT_MERGES[] = "max_concurrent_merges";
+
+// defaults
+static const size_t DEFAULT_REFRESH_INTERVAL = 1000ul;
+static const size_t DEFAULT_COMPILE_KEY_THRESHOLD = 10000ul;
+static const char DEFAULT_KEYVIMERGER_BIN[] = "keyvimerger";
+
+// spinlock wait time if there are to many segments
```
amit-cliqz: typo: to --> too

keyvi/include/keyvi/index/internal/index_auto_config.h
```diff
@@ -0,0 +1,65 @@
...
+namespace keyvi {
+namespace index {
+namespace internal {
+
+class IndexAutoConfig final {
+ public:
+  static size_t MaxSegments() {
+    // increase file descriptors and leave some extra room, to be improved later
+    size_t file_descriptors = keyvi::util::OsUtils::TryIncreaseFileDescriptors();
+    return file_descriptors - 100;
```
amit-cliqz: Magic number 100: As I understand, you don't want to use all the file descriptors. Can you please explain why did you choose 100?

keyvi/include/keyvi/index/internal/index_writer_worker.h
```diff
@@ -209,6 +211,20 @@ class IndexWriterWorker final {
   merge_policy_t merge_policy_;
   util::ActiveObject<IndexPayload> compiler_active_object_;
 
+  void CompileIfThresholdIsHit() {
+    if (++payload_.write_counter_ > payload_.compile_key_threshold_) {
+      compiler_active_object_([](IndexPayload& payload) { Compile(&payload); });
+      payload_.write_counter_ = 0;
+
+      // worst case scenario, to many segments, throttle further writes until we are below the limit
```
amit-cliqz: typo: to --> too

keyvi/include/keyvi/index/internal/index_auto_config.h
hendrikmuhs: I agree, thats extremely vague. The problem is basically that I want to leave some room between the total limit and the setting, so that the client can have some open descriptors as well, e.g. because you create the index from a reading a open file. The other reason: compiling requires 2 open descriptors, so I need some room for that as well.

I think there is no perfect answer. I will improve the documentation. On the other side: Incrementing the file descriptors already helps, on Linux it's upped from 1024 (easy to hit) to 1 million, hitting 1 million segments == file descriptors is quite unlikely.

amit-cliqz: Thanks for the answer, it's clear now.

keyvi/include/keyvi/index/internal/index_writer_worker.h
```diff
@@ -85,7 +90,11 @@ class IndexWriterWorker final {
     boost::filesystem::path index_toc_file_part_;
     internal::IndexSettings settings_;
     std::list<MergeJob> merge_jobs_;
-    size_t max_concurrent_merges_ = 2;
+    size_t max_concurrent_merges_;
+    size_t max_segments_;
+    size_t compile_key_threshold_;
+    size_t index_refresh_interval_;
```
narekgharibyan: Looks like these members are only read after initialization, so we can have them `const` and initialize in constructor initializer list.

keyvi/tests/keyvi/util/single_producer_consumer_ringbuffer_test.cpp
```diff
@@ -69,6 +72,7 @@ BOOST_AUTO_TEST_CASE(SimpleTest) {
 
 BOOST_AUTO_TEST_CASE(SizeTest) {
   SingeProducerSingleConsumerRingBuffer<Item, 11> s;
+  auto now = std::chrono::system_clock::now();
```
narekgharibyan: Looks this var is unused.

keyvi/include/keyvi/util/os_utils.h
```diff
@@ -0,0 +1,60 @@
...
+    struct rlimit limit;
+
+    getrlimit(RLIMIT_NOFILE, &limit);
+    if (limit.rlim_cur != limit.rlim_max) {
+      // try different limits, as platforms behave differently
+      std::vector<size_t> limits{limit.rlim_max, 10000ul, 3200, 1024};
+
+      for (size_t l : limits) {
+        limit.rlim_cur = l;
+        if (setrlimit(RLIMIT_NOFILE, &limit) == 0) {
```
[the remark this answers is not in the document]
hendrikmuhs: I am re-reading the limit now, that seems to me the most simplest and safest solution, because - as I read - setting the limits might be flaky on some platforms.

PR #83, opened 2018-05-15
Added --version option to keyvi cli

narekgharibyan:
- added --version option to keyvi cli
 - moved version into keyvi/__init__.py file
 - bumped version to 0.3.1dev

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -10,9 +10,6 @@ import multiprocessing
 import shutil
 import glob
 import platform
-import re
-import codecs
-
 
 from contextlib import contextmanager
 from os import path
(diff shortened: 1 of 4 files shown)


python/src/py/keyvi/__init__.py
```diff
@@ -17,6 +17,8 @@
 limitations under the License.
 '''
 
+__version__ = '0.3.1dev'
```
hendrikmuhs: I think Pep440 want to have a number after dev, e.g. '0.3.1dev0'

https://www.python.org/dev/peps/pep-0440/

python/setup.py
```diff
@@ -26,6 +29,16 @@
     cpu_count = 1
 
 
+def get_version():
+    here = os.path.abspath(os.path.dirname(__file__))
+    version_file_path = os.path.join(here, 'src/py/keyvi/__init__.py')
+    version_file = codecs.open(version_file_path, 'r').read()
+    version_match = re.search(r"^__version__ = ['\"]([^'\"]*)['\"]", version_file, re.M)
```
hendrikmuhs: what about splitting the version into parts, major, minor, etc. Then we can construct the version any way we need it without requiring regex logic.

If we then add a parameter to setup.py (e.g. `--release`) it could omit the `dev0` suffix for releases. That way we do not need another commit for releases. `--release` could be automatically added by our travis script if run from a tag, that would combine our already working automatic release packaging with this.

python/src/py/keyvi/__init__.py
hendrikmuhs: as said what about:
```
__version_major__ = 0
__version_minor__ = 3
__version_patch__ =  1
__version_dev__ = 0

# to be set externally
__version_dev_mode__ = true 

if __version_dev_mode__: 
    __version__ = "{}.{}.{}dev{}".format(__version_major__, __version_minor__, __version_patch__, __version_dev__)
else:
   __version__ = "{}.{}.{}".format(__version_major__, __version_minor__, __version_patch__)
```

`__version_dev_mode__ ` should then be false for releases. Some little logic would be required to set this. Maybe `setup.py` can write or touch a file for that.

python/setup.py
narekgharibyan: @hendrikmuhs agree! 
It will make development more convenient,  but as you have mentioned that will require some logic to be implemented (which needs to work for `sdist` as well), so I think it's better to add that in a separate PR.

hendrikmuhs: I looked a bit around in other projects. Scipy is doing it similar to my suggestion, they have a `ISRELEASED` bool in their setup.py. The release process seems to be: create branch from master and then they commit a change on the release branch that sets `ISRELEASED` to true and on master they bump the version. Lbnl they create a tag on the release branch.

I think that makes sense, otherwise you have to: commit a change to remove 'dev0', tag the release, create a commit that bumps the version. Creating a release branch isn't a bad choice either.

Having all that said, it seems quite some overhead for a small project, but looks like we become more serious, so why not.

narekgharibyan: I was checking `numpy` and they also have `ISRELEASED` bool. So that seems pretty common.

On the other hand the solution with `--release` was quite convenient, but seems it will be a bit tricky to get it work with `sdist`.

So I'll implement the version with `ISRELEASED` bool.

narekgharibyan:
@hendrikmuhs exactly!

PR #86, opened 2018-06-25
cleanup code style, fix a potential out of bounds bug

hendrikmuhs:
cleanup naming, use C++ limits instead of limits.h. Fix a potential out of bounds access.

Note: The out of bounds problem does not happen in production, at least not until now except for the unit test. In prod the 1st bit vector is always larger than the to second one, nevertheless good to fix it.

### DIFF
--- a/keyvi/include/keyvi/dictionary/fsa/internal/bit_vector.h
+++ b/keyvi/include/keyvi/dictionary/fsa/internal/bit_vector.h
@@ -76,9 +76,9 @@ struct BitVector final {
    */
   template <std::size_t TsizeOther>
   inline void SetVector(const BitVector<TsizeOther>& other, const size_t start_bit) {
-    size_t byte_position = start_bit >> 5;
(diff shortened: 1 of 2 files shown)


keyvi/include/keyvi/dictionary/fsa/internal/bit_vector.h
```diff
@@ -76,20 +76,22 @@ struct BitVector final {
    */
   template <std::size_t TsizeOther>
   inline void SetVector(const BitVector<TsizeOther>& other, const size_t start_bit) {
-    size_t bytePosition = start_bit >> 5;
-    size_t bitPosition = start_bit & 31;
-    size_t write_length = std::min(bits_.size() - bytePosition, other.bits_.size());
+    size_t byte_position = start_bit >> 5;
```
narekgharibyan: As we are touching this code, will be good to make variables `const` along the way.

PR #91, opened 2018-08-02
fix types in stringdistance calculations

hendrikmuhs:
fix type inconsistencies.

This - at least for me - fixes all warnings.


narekgharibyan:
@hendrikmuhs great!
Just one question, why we don't use unsigned type for `max_edit_distance` ?

hendrikmuhs:
The distance matrix is using int32_t which for me makes sense as it should be generic and flexible and allow other distance measures to implement negative scoring.

As max_edit_distance is compared against the score from the distance matrix it can compare the score without the need of a static cast. That's all.

Having that said max_edit_distance is strictly positive, it should never be < 0. We could add a check or use unsigned for the API and internally use signed.

PR #100, opened 2018-09-09
filter some build files from coveralls

hendrikmuhs:
fixes the coveralls filter to ignore some build files


narekgharibyan:
Probably for some reason that results was `NaN`, so it was not taken into account.

screen shot 2018-09-11 at 10 34 40

hendrikmuhs:
yes, still good to get rid of this.

hendrikmuhs:
yes, that's all. Squashing will make sure, it is just 1 commit in the end.

PR #109, opened 2018-11-28
Refactor dictionary properties

### LINKED ISSUE #80: Replace boost property tree with rapidjson
The code uses rapidjson for json parsing but also the json version of boost property tree in other places. It should be possible to replace all boost property trees with rapidjson.

(FWIW: The reason is historical, property trees have been used before all the json support was added)

hendrikmuhs:
re-factor the file properties (header parts of a keyvi file to store entry points, type information etc.) into 2 property classes (for the FSA part and the value part), both for reading and writing.

json parsing/writing now uses rapidjson instead of boost property trees.

fixes #80


hendrikmuhs:
The re-factored code is now writing the same way as property trees, meaning it writes all numbers as strings. So looks a bit ugly but keeps backwards compatibility which is more important. It would be a good idea to fix it, but in that case we should revisit the whole format.

I plan to implement some more tests before this is ready.

keyvi/include/keyvi/dictionary/fsa/generator.h
```diff
@@ -295,7 +290,12 @@ class Generator final {
     }
 
     stream << KEYVI_FILE_MAGIC;
-    WriteHeader(stream);
+
+    keyvi::dictionary::DictionaryProperties p(2, start_state_, number_of_keys_added_, number_of_states_,
```
erik-cliqz: The `2` here refers to the version. I think it makes sense to create a constant somewhere with the current version.

keyvi/include/keyvi/dictionary/fsa/internal/value_store_properties.h
```diff
@@ -0,0 +1,174 @@
...
+    std::string compression;
+    if (value_store_properties.HasMember(COMPRESSION_PROPERTY)) {
+      compression = value_store_properties[COMPRESSION_PROPERTY].GetString();
+    }
+
+    return ValueStoreProperties(offset, size, number_of_values, number_of_unique_values, compression);
+  }
+
+ private:
+  size_t offset_ = 0;
```
erik-cliqz: What does `offset_` refer to?

keyvi/include/keyvi/dictionary/fsa/internal/json_value_store.h
```diff
@@ -310,47 +301,39 @@ class JsonValueStoreAppendMerge final : public JsonValueStoreBase {
   void CloseFeeding() {}
 
   void Write(std::ostream& stream) {
-    boost::property_tree::ptree pt;
-    pt.put("size", std::to_string(values_buffer_size_));
-    pt.put("values", std::to_string(number_of_values_));
-    pt.put("unique_values", std::to_string(number_of_unique_values_));
+    // todo: preserve compression
+    ValueStoreProperties properties(0, values_buffer_size_, number_of_values_, number_of_unique_values_, std::string());
```
erik-cliqz: Try to use either `""` or `std::string()` consistently.

keyvi/include/keyvi/dictionary/fsa/internal/json_value_store.h
```diff
@@ -273,12 +269,10 @@ class JsonValueStoreMerge final : public JsonValueStoreMinimizationBase {
   }
 
   void Write(std::ostream& stream) {
-    boost::property_tree::ptree pt;
-    pt.put("size", std::to_string(values_buffer_size_));
-    pt.put("values", std::to_string(number_of_values_));
-    pt.put("unique_values", std::to_string(number_of_unique_values_));
+    // TODO(hendrik) write compressor
```
erik-cliqz: Just checking: This is not required before merging?

hendrikmuhs: no, I noticed this being broken, but it is broken already. We basically loose information about applied compression when merging 2 files. But that information has no functional impact, it's only meta-information to know whether compression has been enabled when compiling or not.

PR #112, opened 2019-01-30
move build logic into cmake

hendrikmuhs:
work in progress

This change removes duplication of build logic and moves most of the logic into cmake. Cmake writes a configuration file which `setup.py` takes as input to know how to compile/link the library. Apart from that the cmake file has been overhauled to follow best practices.

Notes:
 - only tested on linux, likely breaks on OSX


hendrikmuhs:
The decrease in coverage is due to a change in compiler flags, before: `-O3 --coverage`, after this PR `-O0 --coverage`, `-O3` seems to decrease the number of relevant lines.

PR #114, opened 2019-02-21
add a tiered merge policy

hendrikmuhs:
The tiered merge policy is a smarter selection algorithm for selecting segments to merge. It scores sets of segments while prefering similar sized segments while boosting smaller merges and segments with deletes.

The algorithm is inspired by Lucene's tiered merge policy.

This change makes the tiered merge policy the default


hendrikmuhs:
tested this on a larger scale now: > 400 million keys (on my laptop), looks very promising.

PR #116, opened 2019-07-22
embedable build environment re-factor

hendrikmuhs:
refactor include paths and cmake environment to allow embedding in other projects via cmake

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,5 +1,5 @@
 msgpack-python>=0.5.6
 autowrap>=0.16.0
-pytest
+pytest<5
 twine
 wheel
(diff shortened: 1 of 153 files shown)


python/requirements.txt
```diff
@@ -1,5 +1,5 @@
 msgpack-python>=0.5.6
 autowrap>=0.16.0
-pytest
+pytest<5
```
narekgharibyan: Why do we need this change ?

hendrikmuhs: see https://github.com/pytest-dev/pytest/issues/5519

pytest dropped support for python 3.4

This fixes it for now, we might want to drop python 3.4 as well but did not wanted to do this as part of this PR

PR #120, opened 2019-10-18
update python versions

hendrikmuhs:


hendrikmuhs:
expected to fail as docker image needs to rebuild

PR #122, opened 2019-11-01
add a bulk API for setting key/values (index)

hendrikmuhs:
add a bulk API to set a list of key/values at once

### DIFF
--- a/keyvi/include/keyvi/index/types.h
+++ b/keyvi/include/keyvi/index/types.h
@@ -1,0 +1,43 @@
+//
+// keyvi - A key value store.
+//
+// Copyright 2018 Hendrik Muhs<hendrik.muhs@gmail.com>
+//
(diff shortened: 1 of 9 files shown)


keyvi/include/keyvi/index/types.h
```diff
@@ -0,0 +1,43 @@
...
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace keyvi {
+namespace index {
+
+using key_value_vector_t = std::vector<std::pair<std::string, std::string>>;
+using key_values_t = std::shared_ptr<key_value_vector_t>;
```
narekgharibyan: Just a suggestion, maybe it's better to call this `key_values_ptr_t `  ? so it clear it's a pointer. 

I know it's not this way for other types, but still ...

PR #125, opened 2019-11-23
install openssl via homebrew on osx

hendrikmuhs:
install openssl ourselves to fix osx problem for python 3.4.10


hendrikmuhs:
@narekgharibyan 

good point, it seems better to drop it and e.g. use the slot for 3.8.0 - the docker image for the linux builds has it already

hendrikmuhs:
superseeded by #126

PR #129, opened 2020-01-06
improve mset

hendrikmuhs:
make mset parameter flexible, merge flush and flush async

Note:

the mset change is motivated by protobuf, it allows to use protobuf's map implementation directly


keyvi/include/keyvi/index/index.h
```diff
@@ -92,18 +93,16 @@ class Index final : public internal::BaseIndexReader<internal::IndexWriterWorker
 
   void Set(const std::string& key, const std::string& value) { Payload().Add(key, value); }
 
-  void MSet(const key_values_ptr_t& key_values) { Payload().Add(key_values); }
+  template <typename ContainerType>
+  void MSet(const std::shared_ptr<ContainerType> key_values) {
```
narekgharibyan: Prob you wanted to have a `&` here ?

PR #130, opened 2020-03-09
OSX Travis upgrade xcode to 9.4

hendrikmuhs:
upgrade xcode to 9.4

### DIFF
--- a/travis/build_osx_package.sh
+++ b/travis/build_osx_package.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -ex
+set -x
 
 mkdir build
 cd build
--- a/travis/build_osx_package.sh
+++ b/travis/build_osx_package.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -x
+set -ex
 
 mkdir build
 cd build
--- a/travis/build_osx_package.sh
+++ b/travis/build_osx_package.sh
@@ -4,9 +4,9 @@ set -x
 mkdir build
 cd build
 cmake -DCMAKE_BUILD_TYPE=release -DZLIB_ROOT=/usr/local/opt/zlib ..
-make -j 4
+make -j 4 || travis_terminate 1;
 
-./unit_test_all --log_level=test_suite
+./unit_test_all --log_level=test_suite || travis_terminate 1;
 cd ..
 
 
@@ -20,6 +20,6 @@ brew remove zlib
 brew remove snappy
 
 sudo -H pip install wheelhouse/*.whl
-py.test tests
-py.test integration-tests
+py.test tests || travis_terminate 1;
+py.test integration-tests || travis_terminate 1;
 cd ..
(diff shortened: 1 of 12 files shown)


travis/build_osx_package.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -ex
+set -x
```
narekgharibyan: Don't we want to exit if one of commands in this script returns non-zero exit code ?

hendrikmuhs:
> Great, just curious why `-e` for shell interpreter was removed.

I had massive trouble with this in keyvi-server where I upgraded to xcode 9.4, builds broke with no visible reason until I found out its the `-e`. There are some references to it e.g. https://github.com/travis-ci/docs-travis-ci-com/issues/1672
And in the docs: https://docs.travis-ci.com/user/common-build-problems/#my-build-fails-unexpectedly

(Note that they say it's ok in external scripts, we have separate files, however we `source` them, therefore they are internal although being in an external file)

narekgharibyan:
Hah, I see now. 

I think we `source` them because of these 2 exports:
```
export TMPDIR=/Volumes/ram-disk
export PATH="${HOME}/.pyenv/shims/:/root/.pyenv/bin:${PATH}"
```

Maybe we just need to take them out of these scripts and execute instead of `source` ? 

I'm just worried that if we remove `-e` then we will have builds/tests failures which will not be reported.

hendrikmuhs:
> I'm just worried that if we remove `-e` then we will have builds/tests failures which will not be reported.

Obviously a valid point, osx does not even compile with xcode 9.4 at the moment:
```
In file included from /Users/travis/build/KeyviDev/keyvi/keyvi/bin/keyvi_c/c_api.cpp:29:
/Users/travis/build/KeyviDev/keyvi/keyvi/include/keyvi/dictionary/completion/multiword_completion.h:52:45: error: implicit instantiation of undefined template 'boost::shared_ptr<keyvi::dictionary::Dictionary>'

  explicit MultiWordCompletion(dictionary_t d) : fsa_(d->GetFsa()) {}

                                            ^

/usr/local/include/boost/exception/exception.hpp:15:44: note: template is declared here

namespace boost { template <class T> class shared_ptr; }

                                           ^

In file included from /Users/travis/build/KeyviDev/keyvi/keyvi/bin/keyvi_c/c_api.cpp:30:

/Users/travis/build/KeyviDev/keyvi/keyvi/include/keyvi/dictionary/completion/prefix_completion.h:53:42: error: implicit instantiation of undefined template 'boost::shared_ptr<keyvi::dictionary::Dictionary>'

  explicit PrefixCompletion(dictionary_t d) : fsa_(d->GetFsa()) {}
```

narekgharibyan:
Hah, I believe we can just use `std::shared_ptr` instead, I don't think we have a special reason for using `boost::shared_ptr`, wdyt ?

hendrikmuhs:
> Hah, I believe we can just use `std::shared_ptr` instead, I don't think we have a special reason for using `boost::shared_ptr`, wdyt ?

Fully agree, I somewhat remember that autowrap had problems with `std::shared_ptr` and that's why boost has been used in some places, but I think this is no longer the case.

narekgharibyan:
@hendrikmuhs FYI https://github.com/uweschmitt/autowrap/pull/83

hendrikmuhs:
created #133 and #134 as follop ups

narekgharibyan:
@hendrikmuhs cool, I'll take care of #133

PR #137, opened 2020-04-03
update manylinux docker image base 

narekgharibyan:
- updated manylinux base image to `quay.io/pypa/manylinux2014_x86_64`, more info here: https://www.python.org/dev/peps/pep-0599/
 - dropped python 2.7 support


narekgharibyan:
Sure thing! done in fea2764

PR #142, opened 2020-04-11
Add IntDictionaryCompilerSmallData

narekgharibyan:
- adds `IntDictionaryCompilerSmallData`
 - changes `IntValueStoreBase::value_t` to `uint64_t`


keyvi/include/keyvi/dictionary/fsa/internal/int_value_store.h
```diff
@@ -41,7 +41,7 @@ namespace internal {
 
 class IntValueStoreBase {
  public:
-  typedef uint32_t value_t;
+  typedef uint64_t value_t;
```
hendrikmuhs, line 44: makes sense. This limit is probably from the very beginning when keys could only address 4GB.

Can you also change it for `IntInnerWeightsValueStoreBase`?

narekgharibyan: addressed in eb80c2c

PR #146, opened 2020-04-16
Rust msgpacked value

narekgharibyan:
added functionality to get internal msgpacked on rust end


keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -139,6 +148,16 @@ char* keyvi_match_get_value_as_string(const keyvi_match* match) {
   return std_2_c_string(match->obj_.GetValueAsString());
 }
 
+keyvi_bytes keyvi_match_get_msgpacked_value(const struct keyvi_match* match) {
+  const std::string msgpacked_value = match->obj_.GetMsgPackedValueAsString();
+
+  const size_t data_size = msgpacked_value.size();
+  auto data_ptr = malloc(data_size);
+  memcpy(data_ptr, msgpacked_value.c_str(), data_size);
```
hendrikmuhs, line 156: I am missing some checks:

 - what if msgpacked_value has size 0 (key-only dictionary, empty match)
 - malloc could return NULL on error

narekgharibyan: @hendrikmuhs thx! 
addressed them here ee0f081

PR #147, opened 2020-04-17
Drop global imports 

narekgharibyan:
Apparently we were quite dependent on this :)


hendrikmuhs:
I quickly checked: docs and examples in python(`python/examples`) use  old style imports, too.

narekgharibyan:
@hendrikmuhs addressed in 98aac26 and dae0a95. Hopefully I didn't miss anything.

FWIW, a lot has changed (like python 2-3 upgrade and other stuff), so I believe we should revisit examples separately.

PR #159, opened 2020-06-24
update examples to python 3

hendrikmuhs:
rewrite examples python 3, use .kv instead of .keyvi as file suffix


narekgharibyan:
❤️

PR #168, opened 2020-08-17
Add aarch64 support

vvucetic:
`msse4.2` doesn't exist in `aarch64` architecture. I added exception so that this argument is used in all architectures other than `aarch64`. I didn't want to cover all other flavors of arm because I was not able to test it. 

This compiled and passed all tests in the `aarch64` machine.


rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "keyvi"
-version = "0.4.6"
+version = "0.4.7"
```
narekgharibyan, line 3: @vvucetic please see [release procedure](https://github.com/KeyviDev/keyvi/blob/master/doc/RELEASE_PROCESS.md) keyvi has.

on master branch it always has current dev/unreleased versions, so no need for this version bump here.

narekgharibyan, line 3: @vvucetic please see [release procedure](https://github.com/KeyviDev/keyvi/blob/master/doc/RELEASE_PROCESS.md) keyvi has.

on master branch it always has current dev/unreleased versions, so no need for this version bump here.

vvucetic: Reverted in https://github.com/KeyviDev/keyvi/pull/168/commits/96322e8869cebab2d6fa7d02b47320cce01c1fe2

narekgharibyan:
One possible alternative I see is using command line variables passed with `-D`, and having `msse4.2` enabled by default.
But no strong opinion on this.

@hendrikmuhs prob you are on vacation, so I'll just go ahead and merge this. 
In case you'll have some concerns, we can revisit them later on I believe.

hendrikmuhs:
:+1: Thanks @vvucetic  and @narekgharibyan 

I tried arm, too and was able to compile/index larger amounts of data without a problem.

When I recently looked into improving cmake, I planned to use this: https://github.com/hideo55/CMake-FindSSE/blob/master/FindSSE.cmake This seems to be used by a lot of other projects.

As for arm support in general: The sse4.2 extensions speed up mainly dumping/merging, it makes a huge difference. It bascally does "fast intersection". If I remember correctly its about 8 times faster vs. the vanilla implementation. This code is also used for fuzzy matching afaik.

The other usage of sse is in the compiler and is used to find the first bit set. The vanilla implementation is roughly 50% slower, but looking at the whole compilation, I would guess a difference of 10-20%.

A 3rd usage is in rapidjson, but I doubt this makes a big difference.

Arm has intrinsics, too (neon). At least bit functions are available, but I am not sure about fast intersection, I found some comparison intrinsic support, it would require some investigations to find out whether this can be used to speedup fast intersection.

Anyway, good 1st PR, in the long run I would switch to FindSSE.

PR #172, opened 2020-10-27
rewrite json to msgpack and back and remove 3rdparty dependency

hendrikmuhs:
rewrite of how json is parse to msgpack and back, removing the 3rdparty dependency and replacing it with a simpler implementation

Notes:

- encoding is slightly faster (micro-benchmark)
- decoding avoids the temporary conversion to rapidjson which makes it faster, too
- the removal of the xchange dependency makes it possible to update msgpack-c (tbd in a separate PR)


hendrikmuhs:
the only build failure left is the rust build, which is currently broken for master

narekgharibyan:
@hendrikmuhs I'll look into rust build failure.

keyvi/include/keyvi/util/msgpack_util.h
```diff
@@ -38,69 +40,111 @@
...
+      writer->Uint64(o.via.u64);
+      break;
+
+    case msgpack::type::NEGATIVE_INTEGER:
+      writer->Int64(o.via.i64);
+      break;
+
+    case msgpack::type::FLOAT:
+      writer->Double(o.via.f64);
+      break;
```
narekgharibyan, line 109: In [JsonToMsgPack](https://github.com/KeyviDev/keyvi/pull/172/files#diff-9b644e267a5fc2eb412c40993bfd5740eb1cc692ca8801a1be8c5209dc15fc02R61-R66) we are writing floating point numbers calling 2 different methods: `pack_float()` and `pack_double()` and they have different type codes in `msgpack` format spec.

But in here while reading we have case just for `msgpack::type::FLOAT`, not sure if this is enough or we should have separate case for `msgpack::type::DOUBLE` as well.

keyvi/tests/keyvi/util/json_value_test.cpp
```diff
@@ -23,22 +23,25 @@
...
+  std::string input =
+      "{\"hello\":\"world\",\"t\":true,\"f\":false,\"n\":null,\"i\":123,\"j\":-123,\"pi\":3.1416,\"a\":[1,2,3,4],\"d\":"
+      "{\"k\":\"v\"}}";
+
   std::string encoded = EncodeJsonValue(input);
 
   std::string output = DecodeJsonValue(encoded);
 
-  BOOST_CHECK_EQUAL('"' + input + '"', output);
+  BOOST_CHECK_EQUAL(input, output);
```
narekgharibyan, line 44: will be good to add a test case with `single_precision_float=true`, as we have a special handling for that.

narekgharibyan (approved):
LGTM!

Left a few suggestions/question.

keyvi/tests/keyvi/util/json_value_test.cpp
hendrikmuhs: :+1: I assumed we have this covered, but it seems there is only a very basic test here:

https://github.com/KeyviDev/keyvi/blob/master/python/tests/json/json_dictionary_test.py#L59

I will add some coverage on C++ side.

keyvi/include/keyvi/util/msgpack_util.h
hendrikmuhs: `msgpack::type::DOUBLE` is deprecated, we do not use it (and if, it's an alias for `FLOAT`). Encoding and decoding is not 1:1. That means when reading back there is no difference and json does not care either.

hendrikmuhs: FWIW I think I saw a distinction in newer versions for msgpack-c, will keep that in mind, once I update it, which is what I plan for after merging this PR.

narekgharibyan: Cool, btw master is green now.

PR #174, opened 2020-11-02
disable python coverage for debugging

hendrikmuhs:
coverage reporting is broken after upgrading the docker image(or upgrade of coveralls done as side-effect), just trying to find out why


hendrikmuhs:
for follow up, see #179

PR #177, opened 2020-11-03
Use ubuntu 16.04 for coverage build

narekgharibyan:
Image with new `ubuntu:20.04` doesn't produce consistent coverage report. we need to investigate that separately, but no need to block current PRs. so we can use `ubuntu:16.04` base image temporarily.


hendrikmuhs (approved):
LGTM

PR #178, opened 2020-11-04
Msgpack upgrade

hendrikmuhs:
upgrade msgpack, removes a lot of compiler warnings and should allow us to switch to a newer C++ standard


narekgharibyan (approved):
Very cool!

PR #184, opened 2020-11-20
Add new python versions - part 1

narekgharibyan:
This PR adds new python versions on ubuntu builder.
I'll make a follow-up PR, which will drop old versions, after this PR is merged and dockerhub is updated.

On OS X we are installing python during travis run, so I made those updates in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,24 +49,24 @@ env:
   global:
     - GITHUB_REPO='KeyviDev/keyvi'
 
-matrix:
-  include:
-    - os: osx
-      osx_image: xcode9.4
-      compiler: clang
-      env: BUILD_TYPE=osx PYTHON_VERSION=3.6.9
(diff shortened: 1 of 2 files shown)


hendrikmuhs:
Sounds good, I am not sure what is wrong with osx, this might be due to the Travis Billing changes: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

narekgharibyan:
So I think we can merge this and let dockerhub build new images. So I can create second PR.
Coverage is fine too.

I have already reverted OS X changes. 

Wdyt ?

hendrikmuhs (approved):
LGTM

Lets merge it and wait for an answer from travis support.

Before any new PR I suggest to disable the mac builds until we have a solution.

narekgharibyan:
Disabled Mac builds.

PR #185, opened 2020-11-20
Update python versions - part 2

narekgharibyan:
This PR removes old python versions from Ubuntu builder and adds python3.9 support.
It also updates existing versions to their latest minor versions.


hendrikmuhs (approved):
LGTM!

BTW. This will break https://github.com/KeyviDev/keyvi-server/blob/master/.travis.yml which uses the same docker images. So we need part 3. ;-)
But, no urgency, I can do it myself sometime.

narekgharibyan:
Cool, thanks!

PR #186, opened 2020-11-24
Refactor compiler to use a merge based approach for large dictionaries

### LINKED ISSUE #180: Is this statically linking with tpie?
I was wondering if tpie is linked statically or dynamically witth the rest of keyvi?
Based on https://github.com/KeyviDev/keyvi/blob/c7b523b21054dbb2854d08c50fa67816abe49c5f/CMakeLists.txt#L60 it looks like it is a static link. Am I correct?
Thanks!

### LINKED ISSUE #181: Replace TPIE sorting by partioning and merging fsa's
#180 made my think.

For some reason my vision was to replace the sorting code. I realized this might not be necessary, we have everything we need.

I compared the suggestion I gave in #180 on larger data sets: 
 - creating a keyvi file from scratch, utilizing TPIE
 - creating x small keyvi files (using the "small data compilers" which don't use TPIE sort) and run merger on it

I ran different cases, in summary the merge approach was roughly 20% slower. Note, I did not optimize anything (I used simple python scripts). My merge approach had to copy more data, an improved implementation would avoid that.

The idea is as follows

1. create an in-memory sorter
2. if the in-memory sort buffer hits the threshold, sort the data, create an fsa, persist it, free buffers
3. go to 1.
4. after all data has been processed, sort, create, persist the final chunk
5. merge the fsa's and create the final keyvi file

hendrikmuhs:
Refactor dictionary compiler to use a chunk and merge approach for building larger dictionaries instead of relying on external sort from tpie. This effectively removes the 3rdparty dependency and therefore resolves the questionable linkage to it (removal of the git subtree is not part of this PR, but will be done separately, to not overload this PR).

The refactoring simplifies the usage and better separates the use cases. I split the dictionary compiler into 2 separate ones: 

- dictionary compiler is made for compiling dictionaries standalone, e.g. using keyvicompiler. The abstraction of the sorter got removed, the same compile is used whether you compile a small or large dictionary. Merge based compilation only kicks in, if the data exceeds the (configurable) memory limit. For small dictionaries, the compiler now behaves like the previous "small data" compilers.
- dictionary index compiler replaces the previous "stable insert" mode in dictionary compiler. This compiler is aware of insert order and supports deletes. This special compiler is only needed for keyvi index. This compiler does not overflow to disk, as this isn't required for keyvi index. This simplification also allowed the removal of the counter, instead I use `std::stable_sort` 

### Benchmarking

I did various benchmarks. Runtime is highly data dependent. In worst case the compiler was 40% slower for a large dictionary. Smaller dictionaries are compiled faster as we avoid the tpie overhead. 

Given the goal of removing tpie and the simplifications this seems acceptable. Compilation is still reasonably quick, in practice dictionary compilation is not a bottleneck. With some more tweaks we might be able to speed up compilation.

fixes #181 
fixes #180


keyvi/include/keyvi/dictionary/dictionary_compiler.h
```diff
@@ -250,15 +162,183 @@ class DictionaryCompiler final {
...
   typename GeneratorAdapter::AdapterPtr generator_;
   std::string manifest_;
-  size_t count_ = 0;
+  size_t memory_limit_;
+  size_t memory_estimate_ = 0;
+  size_t chunk_ = 0;
   size_t size_of_keys_ = 0;
-  bool sort_finalized_ = false;
-  bool stable_insert_ = false;
+  size_t parallel_sort_threshold_ = 10000;
```
narekgharibyan, line 174: Seems this value is never changed, should it be made as `static const` ?

keyvi/include/keyvi/dictionary/dictionary_compiler.h
```diff
@@ -250,15 +162,183 @@ class DictionaryCompiler final {
...
+    // disable minimization for faster compile
+    keyvi::util::parameters_t params(params_);
+    params[MINIMIZATION_KEY] = "off";
+    fsa::Generator<keyvi::dictionary::fsa::internal::SparseArrayPersistence<uint16_t>, fsa::internal::NullValueStore,
+                   uint32_t, int32_t>
+        generator(params);
+
+    for (const key_value_t& key_value : key_values_) {
+      TRACE("adding to generator: %s", key_value.key.c_str());
+      generator.Add(std::move(key_value.key), key_value.value);
```
narekgharibyan, line 209: `std::move()` here does make a copy, as  `key_value` is `const`, better to directly pass `key_value.key`, so it can go by ref

keyvi/include/keyvi/dictionary/dictionary_compiler.h
```diff
@@ -250,15 +162,183 @@ class DictionaryCompiler final {
...
+      callback_trigger = 1 + (number_of_items - 1) / 100;
+
+      if (callback_trigger > 100000) {
+        callback_trigger = 100000;
+      }
+
+      for (const key_value_t& key_value : key_values_) {
+        TRACE("adding to generator: %s", key_value.key.c_str());
+
+        generator_->Add(std::move(key_value.key), key_value.value);
```
narekgharibyan, line 245: same here for `std::move`

narekgharibyan (approved):
@hendrikmuhs very cool!

I'm curious how big data you were using for profiling ?

hendrikmuhs:
> @hendrikmuhs very cool!
> 
> I'm curious how big data you were using for profiling ?

I used 13 GB log data(json). To test the different code paths I played with low memory limits, to force the overflow to disk. I know we used much larger files in the past, however this data seemed sufficient to check the PR.

keyvi/include/keyvi/dictionary/dictionary_compiler.h
```diff
@@ -84,13 +69,19 @@ class DictionaryCompiler final {
...
+  explicit DictionaryCompiler(const keyvi::util::parameters_t& params = keyvi::util::parameters_t()) : params_(params) {
+    temporary_directory_ = keyvi::util::mapGetTemporaryPath(params);
+    temporary_directory_ /= boost::filesystem::unique_path("dictionary-fsa-chunks-%%%%-%%%%-%%%%-%%%%");
 
     TRACE("tmp path set to %s", params_[TEMPORARY_PATH_KEY].c_str());
 
-    stable_insert_ = keyvi::util::mapGetBool(params_, STABLE_INSERTS, false);
+    memory_limit_ = keyvi::util::mapGetMemory(params_, MEMORY_LIMIT_KEY, DEFAULT_MEMORY_LIMIT_COMPILER);
+
+    if (memory_limit_ == 0) {
```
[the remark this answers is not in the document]
hendrikmuhs: :+1: you are right, makes no sense, because the `mapGetMemory` handles the default case

PR #187, opened 2020-11-29
use github actions for CI for core, python and style checking

hendrikmuhs:
use github actions for CI, implements builders on linux and macOS for core and python build and a workflow for style checking, applies caching where possible


hendrikmuhs:
for some reason, a direct check-in does not work, will create separate PR's

PR #188, opened 2020-11-29
use github actions for CI - core

hendrikmuhs:
use github actions for CI, implements builders on linux and macOS for core

### DIFF
--- a/.github/workflows/keyvi.yml
+++ b/.github/workflows/keyvi.yml
@@ -1,0 +1,104 @@
+# Build on Linux
+
+name: Build keyvi
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ${{matrix.os}}
+    name:  ${{ matrix.type }} on ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        type: ['Release', 'Debug']
+        os: ['macos-latest', 'ubuntu-latest']
+    env:
+      BOOST_MAJOR: 1
+      BOOST_MINOR: 73
+      BOOST_PATCH: 0
+    steps:
+      - name: install Linux deps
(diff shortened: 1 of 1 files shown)


.github/workflows/keyvi.yml
```diff
@@ -0,0 +1,83 @@
+# Build on Linux
+
+name: Build keyvi
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
```
narekgharibyan, line 9: Maybe we want to build other branches too ? 
For example release branches.

.github/workflows/keyvi.yml
```diff
@@ -0,0 +1,83 @@
...
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libsnappy-dev libzzip-dev zlib1g-dev libboost-all-dev
+          sudo apt-get install ccache
+      - name: install macOS deps
+        if: runner.os == 'macOS'
+        run: |
+          brew update
+          brew install zlib snappy openssl gnu-getopt coreutils boost
```
narekgharibyan, line 36: Why do we install `openssl gnu-getopt coreutils` here ? 
I don't remember us doing it before.

narekgharibyan (approved):
LGTM, just left a few suggestions.

.github/workflows/keyvi.yml
hendrikmuhs: afaik, that requires a separate trigger: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release

That trigger is not based on a branch name, which will have a different name anyway, not sure if you can specify wild cards.

.github/workflows/keyvi.yml
hendrikmuhs: not sure, I took this over from the old definitions afair.

Will give it a try.

.github/workflows/keyvi.yml
narekgharibyan: From [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#on)
```
# Triggered when code is pushed to any branch in a repository
on: push
```
So if we remove `branches: [ master ]` part, then it will start building everything I guess.

PR #192, opened 2020-11-30
update workflow image to ubuntu 20.04

hendrikmuhs:
update ubuntu base image to 20.04 to get a newer boost version, additionally update boost in manylinux docker image

relates #186
relates https://github.com/actions/virtual-environments/issues/1816


narekgharibyan (approved):
Do we want to update [style.yaml](https://github.com/KeyviDev/keyvi/blob/master/.github/workflows/style.yml#L16) as well ?

PR #197, opened 2020-12-06
use ccache-action

hendrikmuhs:
move ccache logic to gh action


narekgharibyan (approved):
Looks quite elegant :)

PR #199, opened 2020-12-09
add sdist and bdist testing to gh workflow

hendrikmuhs:
add sdist and bdist testing to gh workflow


narekgharibyan:
@hendrikmuhs what about dropping `python setup.py build --mode Release` testing in favor of `bdist test` ?

To me it feels like `bdist test` will be sufficient.

hendrikmuhs:
> @hendrikmuhs what about dropping `python setup.py build --mode Release` testing in favor of `bdist test` ?
> 
> To me it feels like `bdist test` will be sufficient (essentially kind of a superset).

I would like to run the standalone `build` at least once, because of the custom code in `setup.py`, however we don't need to run it on every platform/os.

narekgharibyan:
Agree, so basically I wanted to suggest maybe splitting to different workflows, like `sdist.yaml`, `bdist.yaml` and `standalone build.yaml` ?

No strong opinion, on this, but though maybe better to have them separate instead of several if checks in one.

hendrikmuhs:
we have already 18 checks, most time spend in provisioning. Therefore I don't like to split it up distribution building.

`bdist` is now used on all platforms, `sdist` and direct building only on one ubuntu runner.

LBNL I disabled travis, as we agreed that we disable the broken doc build. Packaging is still tbd, therefore I kept the travis script folder.

narekgharibyan (approved):

PR #201, opened 2020-12-12
Bump version and repair pypi upload

hendrikmuhs:
bump version, repair package upload (backport from release-0.5.0 branch) and remove travis subfolder


narekgharibyan (approved):
maybe we don't need to keep `docker/ubuntu-builder/Dockerfile` anymore, wdyt ?

.github/workflows/manylinux.yml
```diff
@@ -44,8 +44,8 @@ jobs:
...
-        uses: pypa/gh-action-pypi-publish@master
-        with:
-          user: __token__
-          password: ${{ secrets.pypi_password }}
-          packages_dir: wheelhouse
+        env:
+          PYPI_PASSWORD: ${{ secrets.pypi_password }}
+        run: |
+          PYBIN=/opt/python/${PYTHON_PATH}/bin
+          ${PYBIN}/python -m twine upload -u __token__ -p "${PYPI_PASSWORD}" python/wheelhouse/*
```
narekgharibyan, line 51: not sure I understand this `__token__` syntax, will it work with `twine upload`

hendrikmuhs: yes, its the new [API token](https://pypi.org/help/#apitoken) way of uploading to pypi.

keyvi 0.5.0 got uploaded this way. What I had to revert was the usage of a github action. The pypi upload action used a docker container which is not supported on macOS and unnecessary complicated. I tried a different action but that one failed for the manylinux image where the python interpreter is not in path. 

Long story short, its not worth using a gh-action for a 1 liner.

hendrikmuhs:
> maybe we don't need to keep `docker/ubuntu-builder/Dockerfile` anymore, wdyt ?

I think its not used anymore, I agree, we can remove it.

.github/workflows/manylinux.yml
narekgharibyan: agree, thx!

PR #204, opened 2021-02-13
add ZipStateTraverser

hendrikmuhs:
The current traverser objects only allow to traverse single dictionaries. However indexes consist of several segments==dictionaries. The ZipStateTraverser wraps a collection of traversers(e.g. a traverser per segment) providing a single traverser. Longer term this will allow fuzzy and completion matching on an index.


hendrikmuhs:
@narekgharibyan I am merging this now, feel free to post-review it.

PR #206, opened 2021-03-13
Fix memory limit calculation for memory limits > 99272 MB

### LINKED ISSUE #205: Advice to build a dictionary of already sorted keys
I have looked through the documentation and issues and it looks like `keyvi` itself sorts the data, for example when I read https://github.com/KeyviDev/keyvi/issues/181. In my case I have two `numpy` arrays, one with 5 billion keys and the another one with the corresponding values. We can sort our 5 billion keys in around 30 minutes using `argsort`. Therefore I wonder what would be the quickest way to build the `keyvi` dictionary from this? Does `keyvi` for example already notice the data is sorted (and does not have to check this again)?

hendrikmuhs:
For memory limits > 99272 MB a bug in memory calculation triggered, resulting in super slow execution and loss of minimization. This change fixes the bug, ensuring the maximum possible configuration is returned and not configuration with `0`. It also allows creating an even bigger hashtable by adding another prime. LBNL the minimum memory limit is checked and an exception is thrown.

fixes #205

### DIFF
--- a/keyvi/include/keyvi/dictionary/fsa/internal/lru_generation_cache.h
+++ b/keyvi/include/keyvi/dictionary/fsa/internal/lru_generation_cache.h
@@ -80,7 +80,7 @@ class LeastRecentlyUsedGenerationsCache final {
    */
   void Add(EntryT key) {
     if (current_generation_->Size() >= size_per_generation_) {
-      MinimizationHash<EntryT> *newGeneration = NULL;
+      MinimizationHash<EntryT> *newGeneration = nullptr;
       if (generations_.size() + 1 == max_number_of_generations_) {
         // remove(free) the first generation
         newGeneration = generations_[0];
@@ -90,7 +90,7 @@ class LeastRecentlyUsedGenerationsCache final {
 
       generations_.push_back(current_generation_);
 
-      if (newGeneration == NULL) {
+      if (newGeneration == nullptr) {
         newGeneration = new MinimizationHash<EntryT>();
       }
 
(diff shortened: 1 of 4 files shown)


keyvi/include/keyvi/dictionary/fsa/internal/lru_generation_cache.h
```diff
@@ -68,21 +66,21 @@ class LeastRecentlyUsedGenerationsCache final {
...
+  LeastRecentlyUsedGenerationsCache &operator=(LeastRecentlyUsedGenerationsCache const &) = delete;
+  LeastRecentlyUsedGenerationsCache(const LeastRecentlyUsedGenerationsCache &that) = delete;
 
   /** Add this object.
    * @param key The key to add
    */
   void Add(EntryT key) {
     if (current_generation_->Size() >= size_per_generation_) {
-      MinimizationHash<EntryT>* newGeneration = NULL;
+      MinimizationHash<EntryT> *newGeneration = NULL;
```
narekgharibyan, line 83: Not related to the PR but, as we are touching this line, maybe we can change `NULL` -> `nullptr` ?
So we have more modern C++-ish code gradually.

keyvi/include/keyvi/dictionary/fsa/internal/minimization_hash.h
```diff
@@ -82,19 +83,28 @@ class MinimizationHash final {
...
     size_t best_fit_generations = 0;
     size_t best_fit_maximum_number_of_items_per_table = 0;
 
+    const size_t min_memory_limit =
+        (hash_size_step_table_[hash_min_size_step_] + (hash_size_step_table_[hash_min_size_step_] >> 2)) * sizeof(T) *
+        min;
+
+    if (min_memory_limit > memory_limit) {
+      throw std::invalid_argument("memory limit too low, for the given parameters the limit must be at least " +
+                                  std::to_string(min_memory_limit));
```
narekgharibyan, line 98: This exception message doesn't say is the minimum in bytes, kilo-bytes or other unit.
I believe it will be better to be explicit here, so users can make a better use of this error message.

narekgharibyan (approved):

PR #207, opened 2021-03-13
Test stylecheck fix

hendrikmuhs:
test only


hendrikmuhs:
separated the fix into #208

PR #213, opened 2021-04-10
fixes for compiling under msvc

hendrikmuhs:
fix some errors under msvcc


hendrikmuhs:
it compiles under msvc now, but linking remains a problem. however I accidentally pushed changes directly instead of using my fork, therefore I am merging it now and re-setup my dev environment.

PR #214, opened 2021-04-18
More msvc related fixes

hendrikmuhs:
fix some more issues found on windows


narekgharibyan:
Very very cool KeyVi is getting on Windows, I'm not using it, but I believe many many folks do :)

PR #220, opened 2021-05-10
allow keys with internal nul chars in rust bindings

gmossessian:
Due to difference between Rust string and C string, it is possible to cause panic by submitting a string with a nul character in the middle to dictionary::get. 

This PR returns empty match instead.


rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "keyvi"
-version = "0.5.3"
+version = "0.5.4"
```
hendrikmuhs, line 3: this shouldn't be part of this PR. 

We bump versions as part of a release, there are some more things to do for that (branching, tagging, ...)

Having that said, we can create a patch release soonish if you need the fix.

hendrikmuhs (commented):
@narekgharibyan is the maintainer of the C and rust binding.

However, I have reservations regarding the approach. The limitation is a bug in the rust binding, keyvi supports zero bytes. Returning an empty match as proposed silences a bug. I would understand if the rust binding errors in some way and this error is handled on client side without causing `panic`.

Would it be possible to provide an additional get method in the C-API that takes a pointer and a length? I saw a container called `keyvi_bytes` in the code used in a different context but basically implementing this approach.

narekgharibyan:
> Therefore I leave it up to you.

Then I'd say no need for a new one, and we can just change existing one.

>That means string and binary handling became much easier and we could actually add support for byte objects, if we have a usecase for it.

Cool, then let's keep rust API part as is for now, and change it in the future if need be.

rust/Cargo.toml
gmossessian: thank you, removed version bump in 5dec183

gmossessian:
Thanks for the discussion @hendrikmuhs and @narekgharibyan, I agree this is a better approach. I'll take a look at the library and re-request review when it's ready :)

@nararekgharibyan `cargo fmt` does not return any errors for me locally, furthermore I cannot see CI, I'm not sure it can be run on this PR without approval as I am a first-time contributor.

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -109,8 +109,12 @@ char* keyvi_dictionary_get_statistics(const keyvi_dictionary* dict) {
   return std_2_c_string(dict->obj_->GetStatistics());
 }
 
-keyvi_match* keyvi_dictionary_get(const keyvi_dictionary* dict, const char* key) {
-  return new keyvi_match(dict->obj_->operator[](key));
+keyvi_match* keyvi_dictionary_get(const keyvi_dictionary* dict, const keyvi_bytes key) {
```
hendrikmuhs, line 112: now that I see the API, I realize we don't need to wrap and unwrap it:

```
keyvi_match* keyvi_dictionary_get(const keyvi_dictionary* dict, const char* key, size_t length)
```

hendrikmuhs (commented):
sorry, it was my suggestion to use keyvi_bytes, actually we don't need it.

(FWIW: its odd that we still create a copy, because the the API requires a `std::string`, but lets ignore that for now, we can switch the dictionary getter to a `string_view`, but that requires C++17)

keyvi/bin/keyvi_c/c_api.cpp
gmossessian: good point! done in 6073fdd

hendrikmuhs (approved):
Thanks! LGTM

gmossessian:
Thank you @hendrikmuhs! I'll wait for @narekgharibyan's final approval as well.

narekgharibyan (approved):
@gmossessian thanks for fixing it. LGTM!

PR #223, opened 2021-05-27
Make more of keyvi safe for rust

gmossessian:
This does three things:

* Remove unsafe calls to `unwrap()` in the rust bindings. This PR makes the choice to handle errors internally by printing to stderr and returning an empty object, but I am open to passing the error handling to the caller. 
* Make the remaining `get_` methods in rust bindings safe for strings with internal nul characters.
* Add a `get_all_prefix_completions` method to access the overloaded C++ method `GetCompletions`.


rust/src/keyvi_string.rs
```diff
@@ -52,6 +52,12 @@ impl Deref for KeyviString {
...
     fn deref<'a>(&'a self) -> &'a str {
         let c_str = unsafe { CStr::from_ptr(self.str_) };
-        c_str.to_str().unwrap()
+        match c_str.to_str() {
+            Ok(s) => s,
+            Err(e) => {
+                eprintln!("Err: {}", e);
+                ""
+            }
+        }
```
narekgharibyan, line 61: As for this part: 
To me it seems a bit wrong if we pass error handling to the caller, as this is an internal thing, and issue is with data that was put in keyvi file in first place, and not being able to return it in second place.

On the other hand we need to understand what kind of data can go here, and what are values we support at all.

Ultimately this calls `Match::GetValueAsString()` from c++, so we prob wanna understand/agree what should we expect from it. 
Only valid `utf-8` strings ? Or it can return any binary data ?

@hendrikmuhs will be good to get your thoughts on this too.

narekgharibyan (changes requested):
@gmossessian thanks for making additional `get_` methods safe!! 
Can you maybe add a test case for them ? 

For removal of 2 unsafe calls I kicked off a discussion, I think it will be good to have a conversation before we merge some change to them.

Also had a comment regarding `keyvi_dictionary_get_all_prefix_completions`.

Thanks again for opening this PR.

rust/src/keyvi_string.rs
hendrikmuhs: > Only valid `utf-8` strings ? Or it can return any binary data ?

I think this returns a valid string, at least that's how I see it implemented right now in various places and what the name lets assume. I would also assume `utf-8`. For binary data I think we have `GetRawValueAsString`.

Having that said, `Match` isn't a great design piece. I consider it legacy, some cleanup/proper re-factoring would be good at some point. But for now, I would say, assume a proper string.

narekgharibyan: And by proper string we mean `utf-8` string which can have `0` char in the middle, right ?

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -118,20 +118,26 @@ keyvi_match_iterator* keyvi_dictionary_get_all_items(const keyvi_dictionary* dic
...
 }
 
-keyvi_match_iterator* keyvi_dictionary_get_fuzzy(const keyvi_dictionary* dict, const char* key,
-                                                 size_t max_edit_distance) {
-  return new keyvi_match_iterator(dict->obj_->GetFuzzy(key, max_edit_distance));
+keyvi_match_iterator* keyvi_dictionary_get_all_prefix_completions(const keyvi_dictionary* dict, const char* key,
+                                                                  const size_t key_len) {
+  PrefixCompletion prefixCompletion(dict->obj_);
+  return new keyvi_match_iterator(prefixCompletion.GetCompletions(std::string(key, key_len)));
+}
```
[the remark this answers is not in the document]
gmossessian: removed in 96750f7

rust/build.rs
```diff
@@ -51,27 +51,27 @@ fn main() {
...
-        .whitelist_function("keyvi_match_get_matched_string")
-        .whitelist_function("keyvi_match_get_msgpacked_value")
-        .whitelist_function("keyvi_match_get_score")
-        .whitelist_function("keyvi_match_get_value_as_string")
-        .whitelist_function("keyvi_match_is_empty")
-        .whitelist_function("keyvi_match_iterator_dereference")
-        .whitelist_function("keyvi_match_iterator_destroy")
-        .whitelist_function("keyvi_match_iterator_empty")
-        .whitelist_function("keyvi_match_iterator_increment")
+        .allowlist_function("keyvi_bytes_destroy")
```
gmossessian, line 54: cargo was giving me warnings that `whitelist` is deprecated.

rust/src/keyvi_string.rs
gmossessian: Any conclusion here? Should we consider making KeyviString dereference to a vec<char>? This would break consumers of the following methods:

[statistics](https://github.com/KeyviDev/keyvi/blob/b6dcf4525ee03b1dc0b19351e36d075b3276b0cb/rust/src/dictionary.rs#L60)
[matched_string](https://github.com/KeyviDev/keyvi/blob/111e8b352555d12fc6430784dde96dd5ae17d6ca/rust/src/keyvi_match.rs#L79)
[get_value_as_string](https://github.com/KeyviDev/keyvi/blob/111e8b352555d12fc6430784dde96dd5ae17d6ca/rust/src/keyvi_match.rs#L52)

hendrikmuhs: The python bindings assume `utf-8` for those, but return byte buffers for e.g. `GetRawValueAsString`.

The problem seems not to be KeyviString, but what creates [it](https://github.com/KeyviDev/keyvi/blob/master/keyvi/bin/keyvi_c/c_api.cpp#L42). This correctly copies the buffer from std::string, however afterwards you don't know the length anymore. The later call to `CStr::from_ptr(self.str_)` truncates the string.

To make this safe, KeyviString should store a length, but that requires all interfaces that call `std_2_c_string` to take a size_t pointer where e.g. `keyvi_dictionary_get_statistics` can store the length:

```
char* keyvi_dictionary_get_statistics(const keyvi_dictionary* dict, size_t* length) {
          std::string stats = dict->obj_->GetStatistics(); 
          *length = stats.size();
          return std_2_c_string(stats);
        }
```

With this KeyviString could be created using the buffer and the length. The Rust string could be created using 
[String::from_raw_parts](https://doc.rust-lang.org/std/string/struct.String.html#method.from_raw_parts).

With these steps we can avoid zero terminated string altogether.

Having that said, this looks like a larger re-factoring and I would not mind doing that as separate PR, which means for this PR to assume `utf-8` without null characters.

narekgharibyan (approved):

PR #237, opened 2021-08-14
check stream for WriteToFile

### LINKED ISSUE #224: No errors from `WriteToFile` when output path contains non existent folder
**Describe the bug**
When you call `WriteToFile` and specify output path with folder that doesn't exists, `WriteToFile` silently do nothing

**To Reproduce**
```python
import os
import keyvi.compiler


c = keyvi.compiler.JsonDictionaryCompiler()
c.Add(b'1', b'123')
c.Compile()

output_path = '/tmp/thispathdoesntexists/test.kv'
c.WriteToFile(output_path)

assert os.path.exists(output_path)  # will fail, because file doesn't exists
```

**Expected behavior**

Exception on `WriteToFile`

hendrikmuhs:
assert broken stream

fixes #224

### DIFF
--- a/keyvi/include/keyvi/dictionary/dictionary_compiler.h
+++ b/keyvi/include/keyvi/dictionary/dictionary_compiler.h
@@ -46,6 +46,7 @@
 #include "keyvi/dictionary/fsa/internal/null_value_store.h"
 #include "keyvi/dictionary/fsa/segment_iterator.h"
 #include "keyvi/util/configuration.h"
+#include "keyvi/util/os_utils.h"
 #include "keyvi/util/serialization_utils.h"
 
 // #define ENABLE_TRACING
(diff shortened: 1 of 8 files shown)


keyvi/include/keyvi/dictionary/dictionary_compiler.h
```diff
@@ -162,6 +162,11 @@ class DictionaryCompiler final {
     }
 
     std::ofstream out_stream(filename, std::ios::binary);
+
+    if (!out_stream.good()) {
+      throw std::invalid_argument("failed to open file");
```
narekgharibyan, line 167: Maybe we can include file-name in the message as well ? 
Also seems we are duplication the message in 4 places, prob we can have some either method or generic message to avoid duplication ?

narekgharibyan (commented):
LGTM, but add 2 small suggestions.

keyvi/include/keyvi/util/os_utils.h
```diff
@@ -64,6 +68,15 @@ class OsUtils {
     return limit.rlim_cur;
 #endif
   }
+
+  static inline std::ofstream OpenFileStream(const std::string& filename) {
```
narekgharibyan, line 72: Maybe it's worth to add _out_ as part of function name to highlight it is opening an out file stream, like `OpenOutFileStream` ?

narekgharibyan (approved):
LGTM. 
@hendrikmuhs thanks for addressing suggestions.

PR #243, opened 2021-12-07
cleanup vint implementation

hendrikmuhs:
cleanup vint implementation, improve vint length performance

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -39,7 +39,7 @@ jobs:
           brew update
           brew install zlib snappy boost
           brew install ccache
-          sudo xcode-select -s "/Applications/Xcode_12.app"
+          sudo xcode-select -s "/Applications/Xcode_12.4.app"
       - name: checkout from git
         uses: actions/checkout@v2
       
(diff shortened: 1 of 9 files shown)


narekgharibyan:
Nice!

narekgharibyan:
Btw, seems more and more libs are supporting M1 chip, chances are I'll switch to M1 sometime soon, so can work on making it compatible with M1 🤞

hendrikmuhs:
> Btw, seems more and more libs are supporting M1 chip, chances are I'll switch to M1 sometime soon, so can work on making it compatible with M1 crossed_fingers

It already works for arm, so I don't expect a lot of problems for M1. What would be nice however:

 - continuous integration, the last time I checked github does not support it. Some 3rdparties have it.
 - the code has some significant optimizations for intel processors (intrinsics). Arm has similar intrinsics, but as support isn't implemented yet, the code uses fallbacks. They are are significantly slower. This applies e.g. to fuzzy matching, but also for dictionary creation.

narekgharibyan:
Yup, I'll try to compile it when I get M1, and we can take it from there.

PR #248, opened 2022-04-01
M1 & Python 10 support

narekgharibyan:
This PR makes it possible to build self contained M1 wheels using Mac with M1 chip. 

Note: So far there is no GH actions M1 runner: https://github.com/actions/virtual-environments/issues/2187

Details:
 - Added (temporary) `build_macosx_wheels.sh` script that can be run locally to create self-contained MacOS wheels
 - Removed workaround of static linking on MacOS and using [delocate](https://github.com/matthew-brett/delocate) package instead
 - Removed limitation of building only `x86_64` on MacOS
 - Added support for Python 3.10
 - Dropped support of Python 3.6 as it reached it's EOL
 - Changed the condition in `CMakeLists.txt` to add `-msse4.2` only when Intel architecture is detected
 - Updated `requirements.txt` and added minimum package versions
 - Fixed GH action build on Python 3.9 runner
 - Removed old `ipython*.sh` scripts


hendrikmuhs (approved):
LGTM

PR #251, opened 2022-06-29
(one more) portable c-char in dictionary.rs

y3ti:
There is one more pointer to fix. I overlooked it in my previous pull request:
https://github.com/KeyviDev/keyvi/pull/250

PS. May you release a new version with all latest changes?

### DIFF
--- a/rust/src/dictionary.rs
+++ b/rust/src/dictionary.rs
@@ -66,7 +66,11 @@ impl Dictionary {
 
     pub fn get(&self, key: &str) -> KeyviMatch {
         let match_ptr = unsafe {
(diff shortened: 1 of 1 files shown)


narekgharibyan (approved):
LGTM! 

But it seems we have an issue with our CI, and the build doesn't pass on Python 3.10 on macOS.

I'm traveling now, and will be back in some 10 days. 

I can make a release by that time.

narekgharibyan:
Just an update here: Now we have many more CI jobs failing, so it seems something has changed, and may take a little while to get all fixed and get the package out. :(

narekgharibyan:
@y3ti `0.5.5` is released.

PR #254, opened 2022-07-19
Build x86 + arm builder images

ivanychev:
* Build `keyvidev/manylinux-builder-x86_64` and `keyvidev/manylinux-builder-aarch64`. The `keyvidev/manylinux-builder` tag remains alias to `keyvidev/manylinux-builder-x86_64`
* Update zlib used in the build since the older URL returns 404


docker/manylinux-builder/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM quay.io/pypa/manylinux2014_x86_64
+ARG base_image
+FROM $base_image
```
narekgharibyan, line 2: What is docker convention for arg naming, I thought that BASE_IMAGE would be more in line what I have seen before.

docker/manylinux-builder/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+docker build . \
+    -f Dockerfile \
+    --build-arg base_image="quay.io/pypa/manylinux2014_x86_64" \
+    -t keyvidev/manylinux-builder && \
+    -t keyvidev/manylinux-builder-x86_64
```
narekgharibyan, line 9: I think we don't need to use aliases, as we can simply change GH actions to use new images, so no need for `-t keyvidev/manylinux-builder`

docker/manylinux-builder/build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+docker build . \
+    -f Dockerfile \
+    --build-arg base_image="quay.io/pypa/manylinux2014_aarch64" \
+    -t keyvidev/manylinux-builder-aarch64
+
+docker push keyvidev/manylinux-builder
+docker push keyvidev/manylinux-builder-x86_64
+docker push keyvidev/manylinux-builder-aarch64
```
narekgharibyan, line 18: Please add an empty line to the EOF.

docker/manylinux-builder/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
narekgharibyan, line 1: Should this file be named `build_and_push.sh` ?

narekgharibyan (changes requested):
Thanks @ivanychev for your contribution.

Before we merge this, I have added a few suggestions.

docker/manylinux-builder/Dockerfile
ivanychev: > What is docker convention for arg naming

Lowercase build args are preferred because this differentiates them from the environment variables, which are always in uppercase.

docker/manylinux-builder/build.sh
ivanychev: Done.

docker/manylinux-builder/build.sh
ivanychev: Sure, done

docker/manylinux-builder/Dockerfile
narekgharibyan: Ok

hendrikmuhs (approved):
Thanks @ivanychev! LGTM

PR #255, opened 2022-07-19
Build keyvi for arm as well

ivanychev:
* Use the `aarch64` and `x86_64` images from #254 to actually build `linux/aarch64` and `linux/x86_64` wheels of `keyvi`
* Docker uses [QEMU](https://www.qemu.org) to run containers with images of different architecture, so there's a `docker/setup-qemu-action@v2`
* The docker images used in the build are now specified in the matrix. That's why there's no `container` entry in the job anymore and we need to actually bring the `addnab/docker-run-action@v3` in order to run different images from the matrix in the job

### DIFF
--- a/.github/workflows/manylinux.yml
+++ b/.github/workflows/manylinux.yml
@@ -48,11 +48,10 @@ jobs:
 
 
       - name: Publish package
-        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
         uses: addnab/docker-run-action@v3
         with:
           image: ${{ matrix.manylinux_image }}
-          options: -v ${{ github.workspace }}:/repo -e PYTHON_PATH=${{ matrix.python_path }} PYPI_PASSWORD=${{ secrets.pypi_password }} PYTHON_PATH=${{ matrix.python_path }}
+          options: -v ${{ github.workspace }}:/repo -e PYTHON_PATH=${{ matrix.python_path }} -e PYPI_PASSWORD=${{ secrets.pypi_password }} -e PYTHON_PATH=${{ matrix.python_path }}
           run: |
             PYBIN=/opt/python/${PYTHON_PATH}/bin
-            ${PYBIN}/python -m twine upload -u __token__ -p "${PYPI_PASSWORD}" /repo/python/wheelhouse/*
+            ${PYBIN}/python -m twine upload --repository testpypi -u __token__ -p "${PYPI_PASSWORD}" /repo/python/wheelhouse/*
--- a/.github/workflows/manylinux.yml
+++ b/.github/workflows/manylinux.yml
@@ -54,4 +54,5 @@ jobs:
           options: -v ${{ github.workspace }}:/repo -e PYTHON_PATH=${{ matrix.python_path }} -e PYPI_PASSWORD=${{ secrets.pypi_password }} -e PYTHON_PATH=${{ matrix.python_path }}
           run: |
             PYBIN=/opt/python/${PYTHON_PATH}/bin
+            ${PYBIN}/pip install twine
             ${PYBIN}/python -m twine upload --repository testpypi -u __token__ -p "${PYPI_PASSWORD}" /repo/python/wheelhouse/*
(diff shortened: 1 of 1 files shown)


.github/workflows/manylinux.yml
```diff
@@ -11,41 +11,46 @@ jobs:
     strategy:
       matrix:
         python_path: [ cp37-cp37m, cp38-cp38, cp39-cp39, cp310-cp310 ]
+        arch_image: [ keyvidev/manylinux-builder-x86_64, keyvidev/manylinux-builder-aarch64 ]
```
narekgharibyan, line 14:
```suggestion
        manylinux_image: [ keyvidev/manylinux-builder-x86_64, keyvidev/manylinux-builder-aarch64 ]
```

.github/workflows/manylinux.yml
```diff
@@ -11,41 +11,46 @@ jobs:
...
 
     runs-on: ubuntu-latest
-    container:
```
narekgharibyan, line 16: Why was this deleted ? 

If we delete this, then I'm afraid the `Publish package` would break as that assumes use of specific python installation and path, that is available inside the manylinux docker container, and not on the host.

narekgharibyan (changes requested):
@ivanychev thanks for your contribution. 

Have added a question regarding package publishing, as in it's current form I'm sure it will break.

.github/workflows/manylinux.yml
ivanychev: Done

.github/workflows/manylinux.yml
ivanychev: I'm afraid I couldn't use top-level ARM container here, that's why I use a custom action

.github/workflows/manylinux.yml
```diff
@@ -11,41 +11,48 @@ jobs:
...
         if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
-        env:
-          PYPI_PASSWORD: ${{ secrets.pypi_password }}
-        run: |
-          PYBIN=/opt/python/${PYTHON_PATH}/bin
-          ${PYBIN}/python -m twine upload -u __token__ -p "${PYPI_PASSWORD}" python/wheelhouse/*
+        uses: addnab/docker-run-action@v3
+        with:
+          image: ${{ matrix.manylinux_image }}
+          options: -v ${{ github.workspace }}:/repo -e PYTHON_PATH=${{ matrix.python_path }} PYPI_PASSWORD=${{ secrets.pypi_password }} PYTHON_PATH=${{ matrix.python_path }}
```
narekgharibyan, line 55:
```suggestion
          options: -v ${{ github.workspace }}:/repo -e PYTHON_PATH=${{ matrix.python_path }} -e PYPI_PASSWORD=${{ secrets.pypi_password }} 
```

Seems you already have `PYTHON_PATH` supplied once, also I think you need `-e` option before each env var.

.github/workflows/manylinux.yml
```diff
@@ -11,41 +11,48 @@ jobs:
...
+            cd ~
+            ${PYBIN}/py.test /repo/python/tests/
+            ${PYBIN}/py.test /repo/python/integration-tests/
 
-          cd ~
-          ${PYBIN}/py.test ${GITHUB_WORKSPACE}/python/tests/
-          ${PYBIN}/py.test ${GITHUB_WORKSPACE}/python/integration-tests/
 
       - name: Publish package
         if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
narekgharibyan, line 51: I think you can test run this if you simply change this condition to evaluate to true ... it will most likely fail on not having permissions to upload to pypi. but at least you'll be sure you get everything to `twine upload` point running.

narekgharibyan (changes requested):
@ivanychev LGTM, but asked to fix the upload part.

Also suggested a way of testing it. 
Would be nice if you could test it, and share the logs here.

.github/workflows/manylinux.yml
ivanychev: Good advice, thanks! I did exactly that

.github/workflows/manylinux.yml
ivanychev: Done

PR #257, opened 2022-07-22
Fix - remove build artifacts between mac os builds

narekgharibyan:
Apparently some build artifacts are not compatible between Python version, like Cython generated code. 

This PR removes build artifacts between different versions, so all wheels are build in one script execution.


hendrikmuhs (approved):
LGTM

PR #264, opened 2023-11-07
update python versions

hendrikmuhs:
update python versions


hendrikmuhs:
I had to remove 3.12 from the manylinux build. I will address it in a separate PR.

PR #267, opened 2024-01-03
refactor Python API to be more pythonic

hendrikmuhs:
This PR introduces a way to change the API's into a more pythonic API. The old naming is still possible, but causes a deprecation message. I rewrote all Dictionary and most of the Match classes(except the attributes). Whenever possible I introduced properties.

I intend to rewrite the remaining API's asap in follow-up PR's.

I appreciate feedback about namings and the (correct?) use of properties.

/CC @iamkhav


iamkhav (approved):
Naming and use of Python's `property` seem reasonable to me.
Thanks!

narekgharibyan (approved):
@hendrikmuhs LGTM! 

I guess we might want to have another pass later on, and names of methods themselves, not just casing ... but this is a good first step.

hendrikmuhs:
@narekgharibyan do you have specific methods names in mind?

I think the `Match` object is still a bit messy and could be simplified, but that entails removal not renaming. Renaming should probably be in sync with C++ changes.

narekgharibyan:
> @narekgharibyan do you have specific method names in mind?

Not yet ... one example that come to my mind is to skip `get_` prefix. 

So instead of `get_keys()` to have `keys()`. 

I guess this is more python dict like ... but no strong opinion yet.

hendrikmuhs:
> Not yet ... one example that come to my mind is to skip get_ prefix.

makes sense. This is a python only addition anyway. I will rename to `items()`, `keys()` and `values()`.

hendrikmuhs:
@narekgharibyan @iamkhav 

I went another pass over the names and changed a few of them. I took the builtin re module as example: 

- `match`, `match_fuzzy`, `match_near` for everything that must match from position `0`
- `search`, `search_tokenized` for everything that can match things in the whole input text.

LBNL, I removed the remaining `get_` , e.g. `statistics`, `manifest`, ...

Let me know what you think.

python/requirements.txt
```diff
@@ -4,3 +4,4 @@ msgpack>=1.0.0
 pytest>=7.1.1
 twine>=4.0.0
 wheel>=0.37.1
+cython>=3.0
```
narekgharibyan, line 7: Seems this file doesn't end with "empty line"

narekgharibyan (approved):
LGTM

PR #272, opened 2024-01-06
Update rapidjson

hendrikmuhs:
update the rapidjson subtree


hendrikmuhs:
It seems that I broke the manylinux on arm docker image. I've pushed a new one. The update I made in #263 actually never made it to docker hub as I assumed that the hub images get recreated automatically. But it turns out that docker images have to be pushed manually. Was that always like this or is it due to changes by license changes by dockerhub? The autosync requires a subscription.

Anyway, the docker image still installs 'snappy-devel', but for some reason cmake does not find 'snappy'. I will create a PR with my latest docker files asap.

@ivanychev @narekgharibyan any ideas?

narekgharibyan (approved):
LGTM!

Will continue convo about Docker builds in https://github.com/KeyviDev/keyvi/pull/273

narekgharibyan:
> But it turns out that docker images have to be pushed manually.

I def remember pushing docker images myself before.

PR #269, opened 2024-01-06
Create python-cibuildwheel.yml

hendrikmuhs:
switch the build system for python to [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/). cibuildwheel makes it easier to build python wheels on different platforms and python versions. The switch hopefully makes our life a little easier and lets us concentrate on the more important things.

With the switch we get:

- builds for mac on arm
- a new target: musllinux
- more pypy targets

In addition to the change I implemented the following additions in this PR:

- use of ccache per platform
- publishing of artefacts as part of the build (for testing PRs by installing the wheel locally)
- various toolchain updates, including enabling dependabot

PR #273, opened 2024-01-08
Update docker images, add musllinux build image and ccache support

hendrikmuhs:
update docker files, more use of parallel building, add ccache and support of musllinux

~Note: There seems to be a problem with the manylinux on arm image. For some reason snappy isn't found although it gets installed.~

FWIW There was a problem with the boost download(server unavailable), I added support for a fallback.

### DIFF
--- a/docker/manylinux-builder/Dockerfile
+++ b/docker/manylinux-builder/Dockerfile
@@ -12,7 +12,7 @@ RUN \
         yum clean all; \
     elif which apk; then \
         apk upgrade --no-cache && \
-        apk add --no-cache bzip2-dev snappy-dev python3-dev hiredis-dev zstd-dev; \
+        apk add --no-cache procps bzip2-dev snappy-dev python3-dev hiredis-dev zstd-dev; \
     else \
         echo "unsupported os"; \
         exit 1; \
--- a/docker/manylinux-builder/Dockerfile
+++ b/docker/manylinux-builder/Dockerfile
@@ -3,12 +3,15 @@ FROM $base_image
 
 RUN \
     if which yum; then \
-        yum -y update; \
-        yum clean all; \
-        yum -y install bzip2-devel snappy-devel python-devel hiredis-devel libzstd-devel; \
+        yum -y update && \
+        yum clean all && \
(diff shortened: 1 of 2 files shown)


hendrikmuhs:
I think I found the problem, `hiredis-devel` and `libzstd-devel` are not available for `aarch64`, that's why the whole line failed. The 2 are only necessary for ccache, but that doesn't built on manylinux for `aarch64` anyway.

narekgharibyan:
How much speed improvement we gain by using `ccache` here ?

narekgharibyan (commented):
How much speed improvement we gain by using `ccache` here ?

hendrikmuhs:
The ccache support is only available together with #269. The ccache effect is available in the logs, e.g. 

```
  /usr/local/bin/ccache -s
  Cacheable calls:     6 /  16 (37.50%)
    Hits:              5 /   6 (83.33%)
      Direct:          4 /   5 (80.00%)
      Preprocessed:    1 /   5 (20.00%)
    Misses:            1 /   6 (16.67%)
  Uncacheable calls:  10 /  16 (62.50%)
  Local storage:
    Cache size (GB): 0.0 / 0.5 ( 1.81%)
    Hits:              5 /   6 (83.33%)
    Misses:            1 /   6 (16.67%)
Save cache using key "ccache-macos-12-x86_64-python-2024-01-08T21:30:18.085Z".
/usr/local/bin/gtar --posix -cf cache.tzst --exclude cache.tzst -P -C /Users/runner/work/keyvi/keyvi --files-from manifest.txt --delay-directory-restore --use-compress-program zstdmt
Cache Size: ~8 MB (8075356 B)
Cache saved successfully
```

I admit, this only tells us that the cache did something, but not the gain. However the numbers should go up, I haven't enabled the other python versions yet. More in #269

narekgharibyan:
I guess it might have been easier to review/merge if we separate changes into smaller parts if possible at all. Like: -
 1. PR that fixes the docker images
 2. PR that adds muslin
 3. PR that adds cache

But this is just a suggestion :)

PR #274, opened 2024-01-09
move to Cibuildwheel workflow

hendrikmuhs:
switch the build system for python to [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/). cibuildwheel makes it easier to build python wheels on different platforms and python versions. The switch hopefully makes our life a little easier and lets us concentrate on the more important things.

With the switch we get:

- builds for mac on arm
- a new target: musllinux
- more pypy targets

In addition to the change I implemented the following additions in this PR:

- use of ccache per platform
- publishing of artefacts as part of the build (for testing PRs by installing the wheel locally)
- various toolchain updates, including enabling dependabot

superseds #269


hendrikmuhs:
The 1st run was really slow, over 3 hours for linux on arm builds. Those run on QEMU (emulator). I've disabled testing on all but the python 3.12 image. The bottleneck seems to be compiling. CCache wasn't active for the 1st run by design, lets see if the 2nd is faster. Unfortunately it only helps for musllinux as manylinux images miss ccache support.

It is possible to speed up CI by splitting the workflow into more tasks, e.g. per python version or maybe `<3.10`, `>= 3.10` or `many` and `musl` (but that only exists on linux).

The other problem with this PR: The mac on arm binaries don't work yet, it seems that the C++ part needs to respect cross compilation, similar to scikit which fixed this problem in https://github.com/scikit-build/scikit-build/pull/530, https://github.com/scikit-build/scikit-build/pull/555, https://github.com/scikit-build/scikit-build/pull/556

python/setup.py
```diff
@@ -177,27 +199,31 @@ def cmake_configure(build_path, build_type, zlib_root, additional_compile_flags)
...
     for ext_m in ext_modules:
         flags = getattr(ext_m, key) + additional_flags
         setattr(ext_m, key, flags)
 
 
 def patch_for_custom_zlib(zlib_root):
     for ext_m in ext_modules:
-        include_dirs = [path.join(zlib_root, "include")] + getattr(ext_m, 'include_dirs')
+        include_dirs = [path.join(zlib_root, "include")] + \
+            getattr(ext_m, 'include_dirs')
```
narekgharibyan, line 223: Honestly I don't like `\` in python code. 
So would ask why was this changed ? The line seems to be 90 chars long. 

also alternative can be to use `( ... + ... )` for long expressions or some temp vars.

python/setup.py
```diff
@@ -118,23 +126,34 @@ def symlink_keyvi():
...
 
     cmake_configure_cmd = 'mkdir -p {}'.format(build_path)
     cmake_configure_cmd += ' && cd {}'.format(build_path)
     cmake_configure_cmd += ' && cmake' \
-                        ' -D CMAKE_CXX_FLAGS="{CXX_FLAGS}"'.format(CXX_FLAGS=CMAKE_CXX_FLAGS.strip())
-    cmake_configure_cmd +=  ' -D CMAKE_BUILD_TYPE={BUILD_TYPE}'.format(BUILD_TYPE=build_type)
+        ' -D CMAKE_CXX_FLAGS="{CXX_FLAGS}"'.format(
+            CXX_FLAGS=CMAKE_CXX_FLAGS.strip())
+    cmake_configure_cmd += ' -D CMAKE_BUILD_TYPE={BUILD_TYPE}'.format(
+        BUILD_TYPE=build_type)
```
narekgharibyan, line 141:
```suggestion
    cmake_configure_cmd += f' -D CMAKE_BUILD_TYPE={build_type}'
```

Maybe better we use f-strings if you want to enforce 80 chars line limit ?

narekgharibyan (approved):
LGTM, just had a few suggestions for setup.py. 

Btw, it seems you are trying to enforce 80 chars line limit ? 
I guess it is somewhat short nowadays, but if you think we should have it, then maybe better we have something like `.editorconfig` ? 
Also seems we have lot of other code, that doesn't comply with it.

fwiw, I had no chance to work with `Cibuildwheel` before, but am glad we are using something more modern, that should make our life better.

hendrikmuhs:
@narekgharibyan No intention to enforce a character limit. I just switched to a new box, used vscode and let it format the file. The limit must be the default as we are missing a configuration for all of the python code. I happily add a config and reformat accordingly. Do you have a config that you consider modern/standard these days?

narekgharibyan:
> What about adding a pyproject.toml and configuring a line length of 100?

Yup sounds good ... fwiw, I didn't play with `pyproject.toml` to have an opinion on it, but I think it doesn't really matter.

hendrikmuhs:
I reformatted setup.py with _ruff_. Unfortunately I had to revert my checkin of `pyproject.toml` as _cibuildwhell_ jumped on it and ignored the `requirements.txt` file

I think we need a renovation of the python code base according to the newest standards, however not as part of this PR.

PR #284, opened 2024-02-04
Expose minimum weight in state traverser for completion dictionaries

hendrikmuhs:
this change introduces a better way to set the minimum weight during traversal. It allows to set the minimum weight during state traversal. This is more flexible than todays `BoundedWeightedStateTraverser`. The change also removes unfinished rewrite called BoundedWeightedStateTraverser2.

Note:

My usecase requires filtering from outside of the prefix completer. The goal of this change in combination with #283 is to implement a functor parameter that lets the user control the minimum weight. I want to control the acceptance of a result as well as setting the minimum weight.


keyvi/include/keyvi/dictionary/fsa/state_traverser.h
```diff
@@ -176,6 +195,28 @@ class StateTraverser final {
   const traversal::TraversalPayload<TransitionT> &GetTraversalPayload() const { return stack_.traversal_stack_payload; }
 };
 
+template <>
+inline void StateTraverser<traversal::WeightedTransition>::PreIncrement() {
+  TRACE("preincrement weighted transition specialization");
+  stack_.traversal_stack_payload.min_weight = min_weight_;
```
hendrikmuhs, line 201: indirection might not be necessary, but this could be set directly in line 217

(the 1st version wrote it differently, but this place is global for the whole traversal)

keyvi/include/keyvi/dictionary/fsa/state_traverser.h
```diff
@@ -166,6 +179,12 @@ class StateTraverser final {
   label_t current_label_;
   bool at_end_;
   traversal::TraversalStack<TransitionT> stack_;
+  // this field only used for weighted traversal, ignored otherwise
+  uint32_t min_weight_ = 0;
+
+  inline void PreIncrement() {}
+
+  inline uint64_t FilterByMinWeight(uint64_t state) { return state; }
```
hendrikmuhs, line 187: todo: add explaination: this is for filtering out states that have already been read, meaning that min_weight has been updated _after_ read.

keyvi/include/keyvi/dictionary/fsa/traversal/weighted_traversal.h
```diff
@@ -52,6 +52,20 @@ static bool WeightedTransitionCompare(const WeightedTransition& a, const Weighte
...
+template <>
+struct TraversalPayload<WeightedTransition> {
+  size_t current_depth;
+  uint32_t min_weight = 0;
+};
+
+template <>
+inline void TraversalState<WeightedTransition>::Add(uint64_t s, uint32_t weight, unsigned char l,
+                                                    TraversalPayload<WeightedTransition>* payload) {
+  if (weight >= payload->min_weight) {
```
hendrikmuhs, line 64: that's the only line causing "overhead" if traversal is done in sorted by weight order without filtering (`min_weight` alwasy `0`). I don't think this deservers another abstraction like we had before. It is rather unlikely that one wants to traverse by weight without filtering.

PR #291, opened 2024-03-05
Re-implement multiword completions with filtering support

hendrikmuhs:
implements multiword completion in the new style, supporting custom filters as recently added for the new prefix completions


narekgharibyan (commented):
did do a full review ... just glanced over and lgtm

PR #292, opened 2024-03-08
Implement multiword completions with fuzzy matching support

hendrikmuhs:
implement a fuzzy version of multiword completion matching. This change also fixes a couple of fuzzy completion matching issues and simplifies the distance matrix calculations.

Note:
  - the distance calculation for cases where fuzzy completions are combined with a character deletion, e.g. completing "finnite state"(note the double "n") to "finite state traverser", was totally broken
  - there was also a case where completions required the last character to be an exact match, not sure where this came from, but this can be handled by the caller


narekgharibyan (approved):
lgtm, glad seeing fixes on this front!

PR #293, opened 2024-03-13
fix weight traversal for long keys. The state traverser did not corre…

hendrikmuhs:
fix weight traversal for long keys. The state traverser did not correctly report the inner weight for the long key optimization. The parent weight must be provided when getting outgoing states.

Fixes issues with completions for long keys


iamkhav (approved):
LGTM.
Hendrik explained to me that previously parent weights were not stored after a certain depth in the graph to save memory. 
With this optimized way as far as I understand, they are just being passed on parent nodes' function calls.

PR #298, opened 2024-04-20
refactor matches to be returned as shared pointers

hendrikmuhs:
refactor matches to be returned as shared pointers with the intention to further modernize the code, remove some quirks with copy/assignment and avoid copying match objects in various places.

This avoids 1 deep-copy for every match in the python extension as well as copies when creating the match iterator.

### DIFF
--- a/keyvi/bin/keyvi_c/c_api.cpp
+++ b/keyvi/bin/keyvi_c/c_api.cpp
@@ -147,15 +147,26 @@ bool keyvi_match_is_empty(const keyvi_match* match) {
 }
 
 double keyvi_match_get_score(const keyvi_match* match) {
+  if (match->obj_ == nullptr) {
+    return 0;
+  }
   return match->obj_->GetScore();
 }
 
 char* keyvi_match_get_value_as_string(const keyvi_match* match) {
+  if (match->obj_ == nullptr) {
+    return std_2_c_string("");
+  }
   return std_2_c_string(match->obj_->GetValueAsString());
 }
 
 keyvi_bytes keyvi_match_get_msgpacked_value(const struct keyvi_match* match) {
   const keyvi_bytes empty_keyvi_bytes{0, nullptr};
+
+  if (match->obj_ == nullptr) {
+    return empty_keyvi_bytes;
+  }
+
   const std::string msgpacked_value = match->obj_->GetMsgPackedValueAsString();
 
   const size_t data_size = msgpacked_value.size();
@@ -172,6 +183,9 @@ keyvi_bytes keyvi_match_get_msgpacked_value(const struct keyvi_match* match) {
 }
 
 char* keyvi_match_get_matched_string(const keyvi_match* match) {
+  if (match->obj_ == nullptr) {
+    return std_2_c_string("");
+  }
   return std_2_c_string(match->obj_->GetMatchedString());
 }
 
--- a/keyvi/bin/keyvi_c/c_api.cpp
+++ b/keyvi/bin/keyvi_c/c_api.cpp
@@ -60,7 +60,7 @@ struct keyvi_match {
 };
 
 struct keyvi_match_iterator {
-  explicit keyvi_match_iterator(const MatchIterator::MatchIteratorPair&& obj)
+  explicit keyvi_match_iterator(const MatchIterator::MatchIteratorPair& obj)
       : current_(obj.begin()), end_(obj.end()) {}
   MatchIterator current_;
   const MatchIterator end_;
--- a/keyvi/bin/keyvi_c/c_api.cpp
+++ b/keyvi/bin/keyvi_c/c_api.cpp
@@ -143,7 +143,7 @@ void keyvi_match_destroy(const keyvi_match* match) {
 }
 
 bool keyvi_match_is_empty(const keyvi_match* match) {
-  return static_cast<bool>(match->obj_);
+  return !match->obj_;
 }
 
 double keyvi_match_get_score(const keyvi_match* match) {
--- a/keyvi/bin/keyvi_c/c_api.cpp
+++ b/keyvi/bin/keyvi_c/c_api.cpp
@@ -33,7 +33,7 @@
 
 using keyvi::dictionary::Dictionary;
 using keyvi::dictionary::dictionary_t;
-using keyvi::dictionary::Match;
+using keyvi::dictionary::match_t;
 using keyvi::dictionary::MatchIterator;
 using keyvi::dictionary::completion::MultiWordCompletion;
 using keyvi::dictionary::completion::PrefixCompletion;
@@ -54,9 +54,9 @@ struct keyvi_dictionary {
 };
 
 struct keyvi_match {
-  explicit keyvi_match(const Match& obj) : obj_(obj) {}
+  explicit keyvi_match(const match_t& obj) : obj_(obj) {}
 
-  Match obj_;
+  match_t obj_;
 };
 
 struct keyvi_match_iterator {
@@ -143,20 +143,20 @@ void keyvi_match_destroy(const keyvi_match* match) {
 }
 
 bool keyvi_match_is_empty(const keyvi_match* match) {
-  return match->obj_.IsEmpty();
+  return match->obj_->IsEmpty();
 }
 
 double keyvi_match_get_score(const keyvi_match* match) {
-  return match->obj_.GetScore();
+  return match->obj_->GetScore();
 }
 
 char* keyvi_match_get_value_as_string(const keyvi_match* match) {
-  return std_2_c_string(match->obj_.GetValueAsString());
+  return std_2_c_string(match->obj_->GetValueAsString());
 }
 
 keyvi_bytes keyvi_match_get_msgpacked_value(const struct keyvi_match* match) {
   const keyvi_bytes empty_keyvi_bytes{0, nullptr};
-  const std::string msgpacked_value = match->obj_.GetMsgPackedValueAsString();
+  const std::string msgpacked_value = match->obj_->GetMsgPackedValueAsString();
 
   const size_t data_size = msgpacked_value.size();
   if (0 == data_size) {
@@ -172,7 +172,7 @@ keyvi_bytes keyvi_match_get_msgpacked_value(const struct keyvi_match* match) {
 }
 
 char* keyvi_match_get_matched_string(const keyvi_match* match) {
-  return std_2_c_string(match->obj_.GetMatchedString());
+  return std_2_c_string(match->obj_->GetMatchedString());
 }
 
 //////////////////////
--- a/keyvi/bin/keyvi_c/c_api.cpp
+++ b/keyvi/bin/keyvi_c/c_api.cpp
@@ -60,10 +60,10 @@ struct keyvi_match {
 };
 
 struct keyvi_match_iterator {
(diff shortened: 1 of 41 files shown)


keyvi/include/keyvi/dictionary/match.h
```diff
@@ -44,9 +45,9 @@ struct Match;
 namespace index {
 namespace internal {
 template <class MatcherT, class DeletedT>
-keyvi::dictionary::Match NextFilteredMatch(const MatcherT&, const DeletedT&);
+std::shared_ptr<keyvi::dictionary::Match> NextFilteredMatch(const MatcherT&, const DeletedT&);
```
gmossessian, line 48: can this be already `match_t` if you move `using match_t = std::shared_ptr<Match>` after the forward declaration?

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -143,20 +143,25 @@ void keyvi_match_destroy(const keyvi_match* match) {
 }
 
 bool keyvi_match_is_empty(const keyvi_match* match) {
-  return match->obj_.IsEmpty();
+  return match->obj_ == nullptr;
 }
 
 double keyvi_match_get_score(const keyvi_match* match) {
-  return match->obj_.GetScore();
+  return match->obj_ == nullptr ? 0 : match->obj_->GetScore();
```
gmossessian, line 150: do you want to just use `operator bool` here for `match_t`? Doing `match->obj_ ?` will check whether the _stored_ pointer is null, whereas I think (may be mistaken) `match->obj_ == nullptr` implicitly creates a shared_ptr wrapping a nullptr in order to compare `lhs.get() == rhs.get()`.

keyvi/include/keyvi/dictionary/completion/forward_backward_completion.h
```diff
@@ -236,10 +236,10 @@ class ForwardBackwardCompletion final {
         std::pop_heap(data->results.begin(), data->results.end(), result_compare());
 
         // de-duplicate
-        while (data->last_result.GetMatchedString() == data->results.back().GetMatchedString()) {
+        while (data->last_result && data->last_result->GetMatchedString() == data->results.back()->GetMatchedString()) {
```
gmossessian, line 239: you have slightly different conventions: [here](https://github.com/KeyviDev/keyvi/pull/298/files#diff-1dbc69e2c71a42075572ad6951771eecba3437f9ab9f4afd3168cc2f7dd8f7f1R150) you explicitly check for `== nullptr` but now you're just using `operator bool` for `last_result`. Since `obj` in the first case and `last_result` in this case are both of type `match_t`, I'm not sure whether there's a reason you're doing it different ways.

keyvi/include/keyvi/dictionary/dictionary.h
```diff
@@ -288,8 +269,8 @@ class Dictionary final {
     MatchIterator current_it = iterators.front();
     iterators.pop();
 
-    auto func = [iterators, current_it]() mutable {
-      while (iterators.size() && (*current_it).IsEmpty()) {
+    auto func = [iterators = std::move(iterators), current_it]() mutable {
+      while (iterators.size() && *current_it == nullptr) {
```
gmossessian, line 273: ` *current_it == nullptr` is another comparison of `match_t` directly to `nullptr`, I believe you could just use the `shared_ptr::operator bool`

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -143,20 +143,25 @@ void keyvi_match_destroy(const keyvi_match* match) {
 }
 
 bool keyvi_match_is_empty(const keyvi_match* match) {
-  return match->obj_.IsEmpty();
+  return static_cast<bool>(match->obj_);
```
gmossessian, line 146: I think you want to `return !match->obj_`

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -143,20 +143,25 @@ void keyvi_match_destroy(const keyvi_match* match) {
...
 }
 
 double keyvi_match_get_score(const keyvi_match* match) {
-  return match->obj_.GetScore();
+  return match->obj_ ? 0 : match->obj_->GetScore();
```
gmossessian, line 150:
```suggestion
  return match->obj_ ? match->obj_->GetScore() : 0;
```

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -172,7 +177,7 @@ keyvi_bytes keyvi_match_get_msgpacked_value(const struct keyvi_match* match) {
 }
 
 char* keyvi_match_get_matched_string(const keyvi_match* match) {
-  return std_2_c_string(match->obj_.GetMatchedString());
+  return match->obj_ ? std_2_c_string("") : std_2_c_string(match->obj_->GetMatchedString());
```
gmossessian, line 180:
```suggestion
  return std_2_c_string(match->obj_ ? match->obj_->GetMatchedString() : "");
```

keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -54,14 +54,14 @@ struct keyvi_dictionary {
...
+  explicit keyvi_match(const match_t& obj) : obj_(obj) {}
 
-  Match obj_;
+  match_t obj_;
 };
 
 struct keyvi_match_iterator {
-  explicit keyvi_match_iterator(const MatchIterator::MatchIteratorPair& obj) : current_(obj.begin()), end_(obj.end()) {}
-
+  explicit keyvi_match_iterator(const MatchIterator::MatchIteratorPair&& obj)
```
narekgharibyan, line 63:
```suggestion
  explicit keyvi_match_iterator(MatchIterator::MatchIteratorPair&& obj)
```

If the second `&` is added to move the iter, then we should drop the `const`.

keyvi/include/keyvi/dictionary/completion/multiword_completion.h
```diff
@@ -150,12 +150,12 @@ class MultiWordCompletion final {
...
           } else {
             TRACE("StateTraverser exhausted.");
-            return Match();
+            return match_t();
           }
         }
       };
 
-      return MatchIterator::MakeIteratorPair(tfunc, first_match);
+      return MatchIterator::MakeIteratorPair(tfunc, std::move(first_match));
```
narekgharibyan, line 158: Not sure if there is a need to do `std::move`, I believe the `shared_ptr` object is not heavy itself.

narekgharibyan (commented):
Had a quick pass, and I see that in most of cases `std::move` is used, which makes me think that almost always we have single ownership instead of shared ownership. 

So maybe we better use `unique_ptr<Match>` to enforce this semantics ? 


Or we really need the `shared_ptr<>` ? 

fwiw, `unique_ptr<>` is more lightweight compared to `shared_ptr<>`

hendrikmuhs:
> Had a quick pass, and I see that in most of cases `std::move` is used, which makes me think that almost always we have single ownership instead of shared ownership.
> 
> So maybe we better use `unique_ptr<Match>` to enforce this semantics ?

We have single ownership within keyvi everywhere. However I think this is more a question for the client that is using it:

- C++: not aware of anyone using it directly
- rust/C: wrapped in another struct, so doesn't matter
- python: wrapped in an object using a `shared_ptr` (autowrap)

> 
> Or we really need the `shared_ptr<>` ?

Conceptually no, `unique_ptr` would do and you can turn(move) a `unique_ptr` into a `shared_ptr` pointer any time if you need it.

However autowrap(python) heavily uses `shared_ptr` and does not support `unique_ptr`. That means extra work is required to make autowrap work again. It doesn't seem like a big [problem](https://github.com/OpenMS/autowrap/blob/master/autowrap/ConversionProvider.py#L2075). I will give it a try to see if there are more obstacles than I see at the moment.

keyvi/bin/keyvi_c/c_api.cpp
hendrikmuhs: it doesn't move, but calls `begin()` and `end()` on it, so this should just be a normal ref.

keyvi/include/keyvi/dictionary/completion/multiword_completion.h
hendrikmuhs: Some research:

- https://stackoverflow.com/questions/41871115/why-would-i-stdmove-an-stdshared-ptr
- https://stackoverflow.com/questions/77786702/how-to-avoid-clang-tidy-warning-when-passing-shared-ptr-to-thread-function


Sounds like `move` should be used.

The 2nd link mentions `clang-tidy`, it is on my todo list to integrate it into our style check.

hendrikmuhs:
I tried `unique_ptr`, it does not work:

```
.../keyvi/include/keyvi/dictionary/match_iterator.h:56:7: note: ‘keyvi::dictionary::MatchIterator::MatchIterator(const keyvi::dictionary::MatchIterator&)’ is implicitly deleted because the default definition would be ill-formed:
   56 | class MatchIterator : public boost::iterator_facade<MatchIterator, match_t const, boost::single_pass_traversal_tag> {
      |       ^~~~~~~~~~~~~
.../keyvi/include/keyvi/dictionary/match_iterator.h:56:7: error: use of deleted function ‘std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) [with _Tp = keyvi::dictionary::Match; _Dp = std::default_delete<keyvi::dictionary::Match>]’
In file included from /usr/include/c++/11/memory:76,
                 from .../keyvi/include/keyvi/dictionary/completion/multiword_completion.h:28,
                 from .../keyvi/bin/keyvi_c/c_api.cpp:30:

```

With other words: the match iterator requires:

`unique_ptr( const unique_ptr& )` - which is by design deleted for `unique_ptr`, but not for `shared_ptr`

The above is needed when accessing the current object in an iterator.

narekgharibyan:
Okie-dokie, thanks @hendrikmuhs for giving it a try ... 

> With other words: the match iterator requires:
unique_ptr( const unique_ptr& ) - which is by design deleted for unique_ptr, but not for shared_ptr

So we are "sharing" the `match_t ` object as far as the `MatchIterator` is concerned ? 

Not sure if we have to, prob we can overcome this obstacle later on if we want to. 

Another alternative idea I had is to not use pointers at all, and just delete copy constructor/assignment  for Match and just let it be moved ... but again this is something for the future. 

Again, thanks for giving it a try.

keyvi/include/keyvi/dictionary/completion/multiword_completion.h
narekgharibyan: Cool, thanks! ❤️

narekgharibyan (approved):

hendrikmuhs:
> Another alternative idea I had is to not use pointers at all, and just delete copy constructor/assignment for Match and just let it be moved ... 

That's what I tried 1st and I couldn't overcome some compile errors. As far as I remember similar issues regarding the match iterator.

narekgharibyan:
Does it stem from `boost::iterator_facade<>` implementation ? 

As if I think about it, theoretically at least we should not need the copy, we should be able to move it ... as it is just a one off data transfer object, or I'm missing something ? 🤔

hendrikmuhs:
I don't remember all the details, `boost::iterator_facade<>1 was definitely part of it.

narekgharibyan:
Okie, thanks. ... will keep this in mind.

PR #300, opened 2024-05-16
python: release the GIL during matching in match iterator

hendrikmuhs:
this change releases the GIL during matching in match iterator. Especially for fuzzy matching the GIL is unnecessarily locked for quite some time, causing python apps to not being able to react.


iamkhav (approved):
LGTM

PR #310, opened 2024-10-20
 create an alternative python binding using pybind11

hendrikmuhs:
Checkout if we can use [pybind11](https://pybind11.readthedocs.io/en/stable/) instead of autowrap/cython. Pybind11 is a much nicer tool to wrap C++, actively maintained and modern. 

As an important starting point I was able to wrap the match iterator. The idea is to add the missing features step by step and then switch the implementation once feature complete.


hendrikmuhs:
close in favor of #315

PR #321, opened 2025-01-19
refactor Python API to be more pythonic - part 2

hendrikmuhs:
This PR is a continuation of #267, it changes more API's to be more consistent with naming in python. The old naming is still possible, but causes a deprecation message.


python/src/addons/ForwardBackwardCompletion.pyx
```diff
@@ -0,0 +1,4 @@
+
+
+    def GetCompletions(self, *args):
+        return call_deprecated_method("GetCompletions", "complete", self.complete, *args)
```
narekgharibyan, line 4: isn't `get_completions` better ?

python/src/addons/Index.pyx
```diff
@@ -92,3 +95,21 @@
             cpp_key_values.get().push_back(cpp_kv)
 
         self.inst.get().MSet(cpp_key_values)
+
+    def MSet(self, *args):
+        return call_deprecated_method("MSet", "mset", self.mset, *args)
```
narekgharibyan, line 100:
```suggestion
        return call_deprecated_method("MSet", "m_set", self.mset, *args)
```
?

Without understanding much :)

narekgharibyan (approved):
lgtm, added 2 minor comments/questions.

python/src/addons/Index.pyx
hendrikmuhs: `mset` is [redis](https://redis.io/docs/latest/commands/mset/). It is basically a bulk endpoint to set multiple key/values with one command. A _bulk_ endpoint.

What about renaming it to `bulk_set`?

python/src/addons/ForwardBackwardCompletion.pyx
hendrikmuhs: :thinking: 

In a previous change I named it like [this](https://keyvidev.github.io/keyvi/python/dictionary.html):

- complete_prefix
- complete_multiword
- complete_fuzzy

avoiding prefixes like `get_`

narekgharibyan: Ah, okie ...

python/src/addons/Index.pyx
narekgharibyan: `bulk_set` is much better IMO.

PR #324, opened 2025-02-10
Zstd compression

hendrikmuhs:
add zstd compression support


narekgharibyan (commented):
lgtm, just seems that the newly written files will be possible to open with new versions of keyvi only ... don't remember do we have some sort of min version requirement somewhere in code ?

hendrikmuhs:
Good point! Yes, we have versioning in place. It should be possible to only bump the version if `zstd` is 
used.

Currently an old version would throw an exception the 1st time a value has to be unpacked. With versioning in place it will not be possible to open a dictionary that uses `zstd` in an old version.

hendrikmuhs:
added versioning, `zstd` now requires a recent keyvi version, trying to load a `zstd` dict in `0.6.3` compared to a `zlib` (both built using this PR):
```
>>> d=keyvi.dictionary.Dictionary("zstd.kv")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "_core.pyx", line 316, in _core.Dictionary.__init__
  File "_core.pyx", line 291, in _core.Dictionary._init_0
ValueError: this version of keyvi file is unsupported
>>> d=keyvi.dictionary.Dictionary("zlib.kv")
>>>
```

PR #338, opened 2025-07-15
Migrate to boost process v2

hendrikmuhs:
Migrate to boost process v2, fixes compile problems on newer versions of boost.

In addition:

- fix autowrap API breakage
- fix doc build
- fix concurrency issue in keyvi index which appeared after boost process update on Mac

### DIFF
--- a/keyvi/tests/keyvi/index/index_limits_test.cpp
+++ b/keyvi/tests/keyvi/index/index_limits_test.cpp
@@ -28,6 +28,8 @@
 #else
 #include <sys/resource.h>
 
+#include <string>
+
 #include <boost/filesystem.hpp>
 #include <boost/test/unit_test.hpp>
 
@@ -63,7 +65,7 @@ BOOST_AUTO_TEST_CASE(filedescriptor_limit) {
   using boost::filesystem::temp_directory_path;
   using boost::filesystem::unique_path;
 
-  size_t old_limit = limit_filedescriptors(20);
+  size_t old_limit = limit_filedescriptors(40);
 
   auto tmp_path = temp_directory_path();
   tmp_path /= unique_path("index-limits-test-temp-index-%%%%-%%%%-%%%%-%%%%");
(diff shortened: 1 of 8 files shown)


keyvi/tests/keyvi/index/index_limits_test.cpp
```diff
@@ -28,6 +28,8 @@
 #else
 #include <sys/resource.h>
 
+#include <string>
```
gmossessian, line 31: why is it needed?

keyvi/tests/keyvi/index/internal/merge_job_test.cpp
```diff
@@ -24,7 +24,9 @@
  */
 
 #include <chrono>  // NOLINT
+#include <string>
 #include <thread>  // NOLINT
+#include <vector>
```
gmossessian, line 29: same here

keyvi/tests/keyvi/index/index_limits_test.cpp
hendrikmuhs: `string` is used on line 39. However it obviously worked before, because it is implicitly included from on of the other includes.

The reason why this pops up now: The code is checked with [`cpplint`](https://github.com/cpplint/cpplint), a static code checker. It produces an error called "include what you use". cpplint _must_ pass as part of the checkstyle workflow run on the PR. checkstyle runs _only_ on files that change. So with other words: Because I made a change to this file, cpplint forced me to cleanup.

keyvi/tests/keyvi/index/internal/merge_job_test.cpp
hendrikmuhs: same answer: `string` and `vector` are used in this file, however this file hasn't been touched since the introduction of cpplint as CI workflow.

gmossessian (approved):

PR #341, opened 2025-07-24
expose loading_stategy in rust

hendrikmuhs:


keyvi/bin/keyvi_c/c_api.cpp
```diff
@@ -103,6 +103,16 @@ keyvi_dictionary* keyvi_create_dictionary(const char* filename) {
   }
 }
 
+keyvi_dictionary* keyvi_create_dictionary_with_loading_strategy(const char* filename,
+                                                                keyvi::dictionary::LoadingStrategy loading_strategy) {
+  try {
+    return new keyvi_dictionary(Dictionary(filename, loading_strategy));
+  } catch (const std::exception& e) {
+    std::cerr << e.what() << '\n';
```
gmossessian, line 111:
```suggestion
    std::cerr << e.what() << std::endl;
```

gmossessian (approved):

keyvi/bin/keyvi_c/c_api.cpp
hendrikmuhs: clang-tidy says [avoid endl](https://clang.llvm.org/extra/clang-tidy/checks/performance/avoid-endl.html).

That's why the other one is already converted to use `\n`. So either I use `\n` or mark the line with `// NOLINT`

PR #357, opened 2025-09-23
Fix sort order of weighted transitions if weights are equal

hendrikmuhs:
use stable sort when sorting weighted transitions, respecting the original order


hendrikmuhs:
See https://github.com/hendrikmuhs/ccache-action/issues/374, brew dropped support for macos-13, aka Ventura.

PR #375, opened 2025-10-21
restore backwards compatibility in string value store

### LINKED ISSUE #333: Breaking change in v0.6.4
**Describe the bug**

There is a change of behavior between v0.6.3 and v0.6.4 that should be considered a breaking change. Previously `dict` values would be returned from `get(...).value`, and v0.6.4 now returns `str` values instead.

**To Reproduce**

Here is a simple script to reproduce of the breaking change:

```python
from keyvi.compiler import StringDictionaryCompiler
from keyvi.dictionary import Dictionary

# Build keyvi file
test = StringDictionaryCompiler()
test.Add("foo", "{}")
test.Compile()
test.WriteToFile("./test.kv")

# Read keyvi file
keyvi_file = Dictionary("./test.kv")
print(type(keyvi_file.get("foo").value))
```

**Expected behavior**
In v0.6.3 the script would print `<class 'dict'>`  (value would be parsed from JSON to a Python `dict`)
While in v0.6.4 the same script now prints `<class 'str'>` (the value is now returned as a Python `str`)

**Additional context**
It's unclear to me if the new behavior is the intended one or not but regardless I believe this is a breaking change that should probably not be shipped as patch release.

hendrikmuhs:
restore backwards compatibility in string value store

fixes #333 
relates #327

### DIFF
--- a/keyvi/include/keyvi/dictionary/fsa/internal/string_value_store.h
+++ b/keyvi/include/keyvi/dictionary/fsa/internal/string_value_store.h
@@ -43,6 +43,7 @@
 #include "keyvi/dictionary/fsa/internal/value_store_properties.h"
 #include "keyvi/dictionary/fsa/internal/value_store_types.h"
 #include "keyvi/util/configuration.h"
+#include "keyvi/util/msgpack_util.h"
 
 // #define ENABLE_TRACING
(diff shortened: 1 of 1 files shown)


hendrikmuhs:
FYI @remusao

